PR #562, opened 2012-02-07
Modified the listener tuple to accept instance protocol value

der4im:
The previous implementation of listener was to accept a tuple with 3 or 4 values e.g. LB port, instance port, protocol and an optional ssl id. I modified it to accept instance protocol as well. This is required if you configure your LB to do HTTPS with client whereas the backend servers are on HTTP.

I have also modified the existing tests cases to reflect the same.


mtdowling:
A non-breaking version of this pull request was merged in https://github.com/boto/boto/commit/b782ce2225a9b17d3a5b3cb94e938dbcb805bfe7. Closing.

PR #631, opened 2012-03-09
Feature: make layer2 update_throughput more scriptable

tmorgan:
have update_throughput handle values greater than 2 times current values, and waiting for table to become active when making changes to read and write units, and return when the table is active. Both of which are required for scripting large changes, or changes to multiple tables (due to subscriber limit of updating one table at a time).


pauloubuntu:
+1 Looking forward to use this feature.

PR #782, opened 2012-05-24
Various fixes for DynamoDB and CloudSearch support

raff:
- when scan or query are called with count=True the result object is not an iterator but an object that contains the counters returned by DynamoDB (count for Count and scanned_count for ScannedCount). The alternative is to have a separate set of methods (scancount and querycount ?)
- invalid format of domain list in list_domains
- SearchResult doesn't store/return facets

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,6 +4,9 @@ boto
 boto 2.5.2
 21-Jun-2012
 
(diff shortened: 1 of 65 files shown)


xethorn:
This pull request can probably be closed? It's been last active in 2012.

PR #1120, opened 2012-11-18
Develop

allyunion:
Pushing my s3bucketsync command into boto.  Works similarly to s3cmd -- accepts credentials from several different types of files.


nagios/README
```diff
@@ -0,0 +1,2 @@
+This is intented as a resource repository for tools that use boto
+and integrate with nagios as nagios checks
```
garnaat: What's this file for?

allyunion: It is my intention to add some nagios integration with cloudwatch using boto... but if that's out of scope of the project, please feel free to reject it.

PR #1214, opened 2012-12-31
Don't use M2Crypto for cloudfront URL signing

tehsmyers:
As far as I can tell, with the exception of some things in contrib, the only reason M2Crypto is a requirement of boto is to be able to sign URLs of CloudFront content. Since I want to be able to sign URLs, but don't want to have to yet another compiled dependency, I took a stab a replacing M2Crypto in cloudfront.Distribution._sign_string with a pure-python equivalent: [rsa](http://pypi.python.org/pypi/rsa).

I tried to keep this scoped in as much as possible to replacing M2Crypto with rsa. I see a couple of unit tests that can be written as well, but they can/should be a different pull.

### DIFF
--- a/boto/cloudfront/distribution.py
+++ b/boto/cloudfront/distribution.py
@@ -656,7 +656,6 @@ class Distribution:
             raise ValueError("You must specify one of private_key_file or private_key_string")
         # If private_key_file is a file, read its contents. Otherwise, open it and then read it
         if isinstance(private_key_file, file):
-            private_key_file.seek(0)
             private_key_string = private_key_file.read()
         elif private_key_file:
             with open(private_key_file, 'r') as file_handle:


pierrei:
I just ran into an issue with the pure Python implementation of `rsa`; it's extremely slow! I'm using an EC2 machine (c3.large) to generate signed URLs and with the rsa implementation it takes ~47 ms to generate _one_ URL. I backported boto to use the M2Crypto version and that one takes about ~2.2ms to generate one URL.

I've [implemented CloudFront support](https://github.com/mathcamp/pypicloud/pull/71) for the [pypicloud](github.com/mathcamp/pypicloud) pypi server, but are noticing really bad performance when listing all versions of a package (similarly to [this page](https://pypi.python.org/simple/boto/)). Every link takes about 50ms to generate so if there are 100 versions of a package, it takes >5 seconds to generate the page.

PR #1273, opened 2013-01-28
Remove fp seek check in s3 set_contents_from_file if fp cannot seek

peterbrook:
Some file-like objects do not support seeking. If the size and md5 of the
underlying data are already known in advance, it is possible and desirable to
read the stream only when we are actually sending the bytes to the storage
service. The current fp.seek calls in this block are simply a sanity check that 
a user didn't pass in an already-advanced fp, and can safely be avoided if the 
fp does not support seeking.


marrotte:
I just ran into this issue and would like to see the feature added.

PR #1390, opened 2013-03-13
Initial GCE Support

shahms:
This pull request picks up and subsumes the similar request from ziyadm adding initial Google Compute Engine support to boto.


nchammas:
@danielgtaylor Will boto potentially consider merging in work like this to support other cloud providers?

PR #1798, opened 2013-10-17
Fix #1792 by retrying 502, 504 HTTP errors.

### LINKED ISSUE #1792: Retry 502 errors
My project has recently seen a rash of transient 502 Bad Gateway and 504 Gateway Timeout errors from the SQS api. Currently the exponential back off mechanism in AWSAuthConnection._mexe is only retrying on 500s and 503s. I believe that 502s and 504s would be appropriate to retry as well. Here's a trivial patch:

```
diff --git a/boto/connection.py b/boto/connection.py
index ed57fc4..18bca5d 100644
--- a/boto/connection.py
+++ b/boto/connection.py
@@ -901,7 +901,7 @@ class AWSAuthConnection(object):
                             boto.log.debug(msg)
                         time.sleep(next_sleep)
                         continue
-                if response.status == 500 or response.status == 503:
+                if response.status in (500, 502, 503, 504):
                     msg = 'Received %d response.  ' % response.status
                     msg += 'Retrying in %3.1f seconds' % next_sleep
                     boto.log.debug(msg)
```

danielgtaylor:
Fixes #1792. Includes tests to make sure those failure codes are retried.

@toastdriven, please sanity check.

### DIFF
--- a/boto/connection.py
+++ b/boto/connection.py
@@ -901,7 +901,7 @@ class AWSAuthConnection(object):
                             boto.log.debug(msg)
                         time.sleep(next_sleep)
                         continue
-                if response.status == 500 or response.status == 503:
+                if response.status in [500, 502, 503, 504]:
                     msg = 'Received %d response.  ' % response.status
                     msg += 'Retrying in %3.1f seconds' % next_sleep
                     boto.log.debug(msg)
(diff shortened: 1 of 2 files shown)


lalinsky:
Sorry for commenting on an old issue, but shouldn't the 502, 504 status codes be also added here?

https://github.com/boto/boto/blob/develop/boto/s3/key.py#L960

PR #1905, opened 2013-12-12
Check no_proxy before changing HTTP request path

mouadino:


mouadino:
No Update for two years, closing ...

PR #2004, opened 2014-01-18
fix https://github.com/boto/boto/pull/1777 for EU location

sashabaranov:


7heo:
I personally prefer https://github.com/boto/boto/pull/2701 to this.

PR #2022, opened 2014-01-23
Add policy parameter support to copy_key().

tpodowd:
- Similar to other API which add a key or a bucket to the S3 server,
  Allow the user to override the default canned-acl which is applied
  to the new key. This default policy override is only used if the
  preserve_acl parameter is False.

### DIFF
--- a/boto/s3/bucket.py
+++ b/boto/s3/bucket.py
@@ -788,7 +788,8 @@ class Bucket(object):
     def copy_key(self, new_key_name, src_bucket_name,
                  src_key_name, metadata=None, src_version_id=None,
                  storage_class='STANDARD', preserve_acl=False,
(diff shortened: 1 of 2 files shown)


tpodowd:
Hi Daniel... Ok that took longer than a week. I just rebased the branch against the latest develop branch and added a new unit test to test things.

    [tpodowd@bertha s3 (add_copy_key_policy)]$ nosetests test_bucket:S3BucketTest.test_copy_key_policy
    .
    Ran 1 test in 3.865s
    
    OK

tpodowd:
@danielgtaylor - I guess also I should mention you by at sign just in case you miss this update.

tpodowd:
@danielgtaylor , @jamesls - any chance of getting this in also.

PR #2030, opened 2014-01-24
Add missing delete_subresource() method to bucket.

tpodowd:
There is a set_subresource() and a get_subresource() but
there was no delete_subresource(). Implement a simple
delete_subresource().

### DIFF
--- a/boto/s3/bucket.py
+++ b/boto/s3/bucket.py
@@ -973,14 +973,14 @@ class Bucket(object):
         :param value: The value of the subresource.
 
         :type key_name: string
-        :param key_name: The key to operate on, or None to operate on the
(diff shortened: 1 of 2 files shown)


tpodowd:
I just rebased this branch to the latest develop. Additionally, I added a missing test case.

    [tpodowd@bertha s3 (delete_subresource)]$ nosetests test_bucket:S3BucketTest.test_delete_subresource
    .
    Ran 1 test in 4.942s
    
    OK

We've been using this patch with Boto for over a year now internally. It should be safe to add and will make adding future subresource commands that bit easier.

tpodowd:
@jamesls - would also like to get this in at some stage. This should be an easy one.

Issue #2043, opened 2014-01-28
Generating a S3 url without authentication fails for IAM roles


larssonp-tmg:
Running key.generate_url always creates the authentication if I run it using an IAM role.

Example code:

```
import boto.s3

conn = boto.s3.connect_to_region('eu-west-1')
bucket = conn.get_bucket('my-bucket')
key = bucket.new_key("%s/%s" % ('testing', 'mytest'))
result = key.set_contents_from_string('my test content')
set_canned = key.set_canned_acl('public-read')
s3url = key.generate_url(0, query_auth=False, force_http=True)
print s3url
```

Making the following changes fixes the issue:

```
-        if self.provider.security_token:
-            headers['x-amz-security-token'] = self.provider.security_token
+        if query_auth:
+            if self.provider.security_token:
+                headers['x-amz-security-token'] = self.provider.security_token
```

ervoiox:
Yes this would be helpful for us too. Please merge.

PR #2044, opened 2014-01-28
check whether query_auth is true before adding x-amz-security-token when run using an IAM role

pjelar:
Raised in ticket:
https://github.com/boto/boto/issues/2043


dwick:
:+1:

PR #2081, opened 2014-02-06
Extend read() method to support no-args and negative sizes (meaning readall)

rivimey:
This patch updates the read() method in the s3 transfer code to support read() as well as read(n). I wanted to be able to use the python zipfile code, which calls both forms, and the existing code fails with a syntax error when used this way.

### DIFF
--- a/tests/unit/s3/test_keyfile.py
+++ b/tests/unit/s3/test_keyfile.py
@@ -42,6 +42,8 @@ class KeyfileTest(unittest.TestCase):
 
     def testReadFull(self):
         self.assertEqual(self.keyfile.read(len(self.contents)), self.contents)
+        self.assertEqual(self.keyfile.read(), self.contents)
+        self.assertEqual(self.keyfile.read(-1), self.contents)
 
     def testReadPartial(self):
         self.assertEqual(self.keyfile.read(5), self.contents[:5])


rivimey:
@danielgtaylor would you be able to have a look at merging this?

boto/s3/keyfile.py
```diff
@@ -84,9 +84,17 @@ def seek(self, pos, whence=os.SEEK_SET):
 
     self.location = pos
 
-  def read(self, size):
-    self.location += size
-    return self.key.read(size)
+  def read(self,*args):
```
signalpillar: as I understand it would be better to use named parameter here instead of vararg. 

```
def read(self, size=-1):
```

2) Could you please document touched method

```
Read at most size bytes from the file (less if the read hits EOF before obtaining size bytes).

:param size: Size of the content to read in bytes. If the size argument is negative or omitted, read all data until EOF is reached. 
:type size: int
:return: .... `None` is returned when EOF is encountered.
:rtype: ?
```

boto/s3/keyfile.py
```diff
@@ -84,9 +84,17 @@ def seek(self, pos, whence=os.SEEK_SET):
...
+    if len(args) == 0 or (len(args) == 1 and args[0] < 0):
+      rv = self.key.read()
+      if rv != None:
+	self.location += len(rv)
+    else:
```
signalpillar: Not sure this branching is required. If `key` follows `file` protocol we can pass negative values too. 
So the resulted should be like

```
result = self.key.read(size)
if result is not None:
    self.location  += len(result)
return result
```

boto/s3/keyfile.py
```diff
@@ -84,9 +84,17 @@ def seek(self, pos, whence=os.SEEK_SET):
 
     self.location = pos
 
-  def read(self, size):
-    self.location += size
-    return self.key.read(size)
+  def read(self,*args):
+    if len(args) == 0 or (len(args) == 1 and args[0] < 0):
+      rv = self.key.read()
+      if rv != None:
```
signalpillar: PEP8: E711 (^)	comparison to None should be ‘if cond is None:’

rivimey:
Thank you for the review. However, I think I will bow out... not pique, but simply that I'm no python expert and there are issues here that go beyond me.

My sole intention in making this change was to enable the code to be used by the python zlib wrapper, which in turn assumes a standard python lib compatible 'read' method.

PR #2100, opened 2014-02-13
Allow security_token to be configured in config profiles

willthames:
Enable the use of security_token with profile sections in
boto configuration file.

Before this change, aws_security_token in profiles is ignored
Afterwards, the aws_security_token value is used

Tests included.


cornfeedhobo:
lol, do you think this will ever get reviewed?

PR #2169, opened 2014-03-18
Fixes layer1 modify_db_parameter_group

mattlong:
modify_db_parameter_group is currently broken, suggesting no one has ever tried to use the method before. Now I'm very curious how much of Boto has actually been tested against AWS...this greatly reduces my confidence in Boto :(

As stated in the documentation, the function should only expect values
for ParameterName, ParameterValue, and ApplyMethod instead of all the
fields supplied when reading parameters from RDS.

### DIFF
--- a/boto/rds2/layer1.py
+++ b/boto/rds2/layer1.py
@@ -2890,7 +2890,7 @@ class RDSConnection(AWSQueryConnection):
         self.build_complex_list_params(
             params, parameters,
             'Parameters.member',
-            ('ParameterName', 'ParameterValue', 'Description', 'Source', 'ApplyType', 'DataType', 'AllowedValues', 'IsModifiable', 'MinimumEngineVersion', 'ApplyMethod'))
+            ('ParameterName', 'ParameterValue', 'ApplyMethod'))
         return self._make_request(
             action='ModifyDBParameterGroup',
             verb='POST',


claudiupopescu:
Instead of deleting all other values you can simply rearrange it to:
```
('ParameterName', 'ParameterValue', 'ApplyMethod', 'Description', 'Source', 'ApplyType', 'DataType', 'AllowedValues', 'IsModifiable', 'MinimumEngineVersion')
```

This is a worrisome since it was not fixed in almost an year.

ivanistheone:
It's disappointing to see such a bug is still around. Not cool. I'm going to write an angry article in the NYTimes and smear you now ;)

For today this horrible hack will have to do:

```
# (            'ParameterName',        'ParameterValue', 
# 'Description', 'Source', 'ApplyType', 'DataType', 'AllowedValues', 'IsModifiable', 'MinimumEngineVersion', 
#                                                           'ApplyMethod'))
param_tuple = ('group_concat_max_len',  153661, 
   None,          None,     None,        None,       None,            'true',        None,
                                                             'pending-reboot')
rds2.modify_db_parameter_group('prod-db-params', [param_tuple])
```

What is the currently recommended approach for scripting RDS? use the older `rds` or switch to `boto3` for RDS stuff? I'm happy with boto2 for everything else...

Issue #1536, opened 2013-06-05
Enhance support to S3 API for parallel upload and files bigger than 5GB


anthonygerrard:
The Amazon S3 Java API directly supports parallel uploads to Amazon S3 and files greater than 5GB via the TransferManager class:

```
TransferManager tm = new TransferManager(s3, 
    (ThreadPoolExecutor) Executors.newFixedThreadPool(threads));
tm.upload(request).waitForCompletion();
```

It would nice if this was as easy to achieve with boto.

According to this post http://bcbio.wordpress.com/2011/04/10/parallel-upload-to-amazon-s3-with-python-boto-and-multiprocessing/  you can do this with boto but I think it would be a useful addition for it to be directly supported.

PR #2256, opened 2014-05-07
Concurrent S3 Uploader

indemand-dev:
Although the S3 Boto interface supports a multipart upload, the multipart upload is not done concurrently.

This pull request adds a concurrent multipart upload to S3 Boto. Since the task of uploading a file to S3 is IO bound especially over high latency links, breaking the upload into multiple parts and uploading them concurrenlty provides a big performance boost. On my workstation which is 7ms RTT from US-East, transfer speeds for a 500MB file increased by 2-3x using 10 threads.

There are many methods available for providing similar functionality, e.g. http://www.elastician.com/2010/12/s3-multipart-upload-in-boto.html and http://bcbio.wordpress.com/2011/04/10/parallel-upload-to-amazon-s3-with-python-boto-and-multiprocessing/ However, in each case the source file is split into smaller parts and each part uploaded separately. This technique introduces multiple problems:
1. It only works on operating systems that natively have the split command
2. The splitting process generates additional local read and write IO
3. The additional split files consume additional local disk space
4. There is no easy way to clean up the split files after a failure, e.g. power failure

A better solution is to use the seek() method on a file pointer to upload parts of the file. The file is logically chunked into parts, the part pointers are added to a worker queue, and each worker uploads part of the file. When all parts are uploaded, the multipart upload is closed. No additional local files are generated.

Here is a Github Gist with sample code for using the ConcurrentUploader https://gist.github.com/mraposa/a4fb2d4a526ac5b2a65a 

I tried to remain as consistent as possible with the Glacier.concurrent module. In fact most of the code is an exact derivation from that module.

If this pull request is accepted, it may be possible to write a concurrent downloader as well. S3 supports get range requests which is similar to what is used with the Glacier Concurrent Downloader.

FYI – I wrote the original Glacier Concurrent Downloader. Pull request #1123. There have been some very nice improvements made since that original pull request.

### DIFF
--- a/boto/s3/concurrent.py
+++ b/boto/s3/concurrent.py
@@ -47,10 +47,6 @@ class UploadConcurrentS3Error(ConcurrentS3Error):
     pass
 
 
-class DownloadConcurrentS3Error(ConcurrentS3Error):
-    pass
-
-
 class ConcurrentTransferer(object):
     def __init__(self, part_size=DEFAULT_PART_SIZE, num_threads=DEFAULT_NUM_THREADS):
         if part_size < 5 * 1024 * 1024:


AgentME:
If this were finished and merged, could it be made the default strategy for uploading files over 5gb? Currently, boto tries to send the file in a single PUT which Amazon doesn't support for that size. Requiring applications to opt in to a different upload method for large files seems like the library is unnecessarily exposing an implementation detail. (This issue seems to be commonly run into: [duplicity](https://answers.launchpad.net/duplicity/+question/253609), [boto-rsync](https://www.cryptobells.com/boto-rsync-limitations-and-workarounds/), [some discussion](https://groups.google.com/forum/#!topic/boto-users/CM1eZFgk-JA), and the idea of making 5gb uploads just work is issue #1536.)

PR #2270, opened 2014-05-15
Handle UnprocessedKeys properly in table.py

### LINKED ISSUE #2214: UnprocessedKeys are dropped in table.py (major bug)
In boto/dynamodb2/table.py:

```
raw_unproccessed = raw_results.get('UnprocessedKeys', {})
for raw_key in raw_unproccessed.get('Keys', []):
```

This won't work because @raw_unprocessed is a dictionary keyed by table name.  The code is assuming the @raw_results looks like:

```
{ "UnprocessedKeys": { "Keys": ... } }
```

But it actually looks like (see http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchGetItem.html )

```
{ "UnprocessedKeys": { "MyTableName1": { "Keys": ...} ... } }
```

As a result of this bug, any unprocessed keys will not be retried, and so not all keys are returned via batch_get().  In my case, when reading 80K rows, in batches of about 20 or 30, I would get back anywhere between 60K and 79K of the 80K total rows, non-deterministically.

I changed the 2nd line of code to

```
   for table_name, table_unprocessed in raw_unproccessed.iteritems():
        for raw_key in table_unprocessed['Keys']:
```

and that fixed the bug.  However, I don't understand all the places dynamodb2/table.py is used, so I don't know if this is the right fix for all.

dustinboswell:
Fixes https://github.com/boto/boto/issues/2214

### DIFF
--- a/boto/dynamodb2/table.py
+++ b/boto/dynamodb2/table.py
@@ -1308,13 +1308,15 @@ class Table(object):
 
         raw_unproccessed = raw_results.get('UnprocessedKeys', {})
 
-        for raw_key in raw_unproccessed.get('Keys', []):
-            py_key = {}
+        for table_name, table_unprocessed in raw_unproccessed.iteritems():
+            # We assume table_name == self.table_name
+            for raw_key in table_unprocessed['Keys']:
+                py_key = {}
 
-            for key, value in raw_key.items():
-                py_key[key] = self._dynamizer.decode(value)
+                for key, value in raw_key.items():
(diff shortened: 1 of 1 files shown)


dcosson:
@danielgtaylor does this still just need a unit test? This seems like a pretty crazy important issue, where boto just silently stops returning any data when getting throttled.

Should I be using boto3 instead? It doesn't look like batch_get or batch_get_item exist in boto3 yet.

PR #2292, opened 2014-05-27
Add support for shared credentials file

danielgtaylor:
Similar to https://github.com/boto/botocore/commit/0c16395e3b024f1b1cb43142f407884ea57fed83

Loads the config in two stages now (normal config, then shared credentials), introduces the `AWS_PROFILE` env var and adds a few new tests to cover possible cases. The loading order from highest to lowest precedence:
1. Directly passed from code
2. Environment variables for key/secret
3. Environment variables for profile
4. Shared credential file explicit profile
5. Shared credential file default profile
6. Config file explicit profile
7. Config file `Credentials` section

A profile passed from code overrides any set in an environment variable.

@jamesls please take a look.

### DIFF
--- a/boto/provider.py
+++ b/boto/provider.py
@@ -280,7 +280,7 @@ class Provider(object):
                 boto.log.debug("Using access key found in shared credential "
                                "file for profile %s." % profile_name)
             elif config.has_option("profile %s" % profile_name,
-                 access_key_name):
+                                   access_key_name):
                 self.access_key = config.get("profile %s" % profile_name,
                                              access_key_name)
                 boto.log.debug("Using access key found in config file: "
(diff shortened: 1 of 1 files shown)


vlcinsky:
I like this feature, really.
Anyway, I think, that the order of preferences is wrong:

1. Directly passed from code
2. Environment variables for key/secret
3. Environment variables for profile
4. Shared credential file explicit profile
5. Shared credential file **default profile**
6. Config file **explicit profile**
7. Config file Credentials section

**Why is any sort of default preferred over explicit profile?**

I would propose swapping 5 and 6, so the order of preference would be

1. Directly passed from code
2. Environment variables for key/secret
3. Environment variables for profile
4. Shared credential file explicit profile
6. Config file **explicit profile**
5. Shared credential file **default profile**
7. Config file Credentials section

What do you think about it?

PR #2307, opened 2014-06-03
Removed the improper setting of a default value for CloudSearch int indexes

josefdlange:
DefaultValue was being set both inside the always-run block, and of course in the block in which the value to set is checked for existence. Removed the former so that empty DefaultValue does not make it into a request (this causes the request to fail)

### DIFF
--- a/boto/cloudsearch2/domain.py
+++ b/boto/cloudsearch2/domain.py
@@ -340,7 +340,6 @@ class Domain(object):
                     ','.join(source_field)
         elif field_type == 'int':
             index['IntOptions'] = {
-                'DefaultValue': default,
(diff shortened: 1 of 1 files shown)


josefdlange:
Totally forgot about and dropped the ball on this. Is this PR even still relevant or should we throw it out? I'd be glad to work out the tests now.

josefdlange:
Still ball-dropping on this. I believe it's so far behind (and likely is already superseded by another fix, though I couldn't find evidence at first glance) that I believe we can abandon it. I'm going to close this, if you think it's still useful and valid let me know and I can work it out. Haven't touched CloudSearch in over a year since leaving a team within Amazon.

PR #2320, opened 2014-06-09
Allow provider to be passed down in connect calls where it currently cannot be.

prufrax:
Ensure that provider can be passed down to base class when connecting to services, so we can override the default provider class.

The base AWSAuthConnection class will already optionally a provider - and will accept an object derived from Provider rather than just a provider type string like 'aws'.

So if the provider is passable down in all classes derived from AWSAuthConnection, it is possible to derive a class from Provider to do e.g. automatic renewal of creds obtained from boto.sts.assume_role() by overriding the mechanism used to renew creds retrieved from instance metadata, and then pass an instance of this in to be used with a connection to e.g. sdb or dynamodb:

foo = AssumedRoleProvider('arn:aws:iam::someaccount:role/somecrossaccountrole')
sdb_conn = boto.sdb.connect_to_region('us-east-1', provider=foo)

This allows code to be used with same-account (instance) roles or cross-account roles (that need to be assumed) with minimal change.

### DIFF
--- a/tests/unit/ec2/elb/test_connection.py
+++ b/tests/unit/ec2/elb/test_connection.py
@@ -1,5 +1,5 @@
 
-from tests.unit import MockServiceProviderTestCas
+from tests.unit import MockServiceProviderTestCase
 from boto.ec2.elb import ELBConnection
 
 


prufrax:
I finally got enough roundtuits to get round to adding some tests to check the provider override, and pulled my fork up to current boto master so it'll auto-merge.

prufrax:
should have probably been forked on master rather than develop to start with, sorry...

dimosch:
Hi,
Any  idea of when should we expect this to be released ?

devinlundberg:
+1 Any updates on this or anything I can do to help this get merged?

yiyan-wish:
Have some legacy code, which using boto, and will be migrate to eks.
This feature can help us fix the aws credentials problem.

PR #2330, opened 2014-06-17
In Table.describe(), try also to update the global indexes information.

ndenev:
Without this I'm unable to perform a query using global secondary index on a table that has single hash key as primary index as boto is erroring with "QueryError: You must specify more than one key to filter on."


pauloubuntu:
@danielgtaylor , @ndenev 

There is no need to merge this PR since it's trying to solve the same issue the PR #2829 which has just been merged by you. You may close this one if you want.

Regards

pauloubuntu:
Ping @danielgtaylor  :)

PR #2337, opened 2014-06-19
Make sure integer values in rds parameter are not based on variables

jimi-c:
Some RDS parameter group values may be based on variables. For example,
by default the innodb_buffer_pool_size parameter has a default value of
'{DBInstanceClassMemory*3/4}' when the group is created. Trying to even
fetch the value of this parameter results in a traceback:

  File "../site-packages/boto/rds/parametergroup.py", line 184, in get_value
    self._set_integer_value(self._value)
  File "../site-packages/boto/rds/parametergroup.py", line 146, in _set_integer_value
    value = int(value)
ValueError: invalid literal for int() with base 10: '{DBInstanceClassMemory/12582880}'

This patch corrects this by checking if the value is a basestring, and
only converting it to an integer if it does not start with a '{'.

### DIFF
--- a/boto/rds/parametergroup.py
+++ b/boto/rds/parametergroup.py
@@ -137,13 +137,17 @@ class Parameter(object):
             raise ValueError('value must be of type str')
         if self.allowed_values:
             choices = self.allowed_values.split(',')
-            if value not in choices:
+            if value not in choices and not value.startswith('{'):
                 raise ValueError('value must be in %s' % self.allowed_values)
         self._value = value
 
     def _set_integer_value(self, value):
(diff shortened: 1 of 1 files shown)


sayap:
We just encountered the same error when using ansible `rds_param_group` module. @jimi-c Maybe you can do a close-and-reopen to trigger another travis build?

danvaida:
@jimi-c any updates on this one? I'm currently researching on the implications of submitting a PR for the `rds_param_group` module to allow setting of dynamic vars in the format of `{DBInstanceClassMemory/16}`.

PR #2347, opened 2014-06-26
Correct unicode conversion error when dumping SQS queue with corrupted messages

filipenf:


filipenf:
SQS code has evolved a lot since my original PR and now unicode is being treated correctly, as of (https://github.com/boto/boto/commit/f098b62c2bbaed171f74fc1f64302dddd607f789). So I believe this one can be closed.

Issue #2050, opened 2014-01-30
S3: setting Content-Language to None causes S3ResponseError


xiongchiamiov:
I'm writing a script to fix the expire headers on our images in S3, as well as modify a few other things while we're at it.  The code looks something like this:

```
for bucketName in bucketsWithImages:
    bucket = s3.get_bucket(bucketName)
    for image in bucket.list():
        image.set_metadata('foo', 'bar')
        image.copy(image.bucket.name, image.name, image.metadata, preserve_acl=True)
```

There are a number of headers that are filtered out of the `metadata` attribute, but are required to be there to not have them lost in the copy operation.  That's a long story for another time, but the result is that I'm calling `image.set_metadata('Content-Language', image.content_language)`.

While testing on a certain subset of images, I kept getting an `S3ResponseError` wrapping a `SignatureDoesNotMatch` error from S3:

```
Traceback (most recent call last):
  File "/home/jpearson/Code/Exec/fixImageCacheHeaders.py", line 78, in <module>
    preserve_acl=True)
  File "/usr/lib/python2.6/site-packages/boto/s3/key.py", line 474, in copy
    encrypt_key=encrypt_key)
  File "/usr/lib/python2.6/site-packages/boto/s3/bucket.py", line 762, in copy_key
    response.reason, body)
boto.exception.S3ResponseError: S3ResponseError: 403 Forbidden
<Error><Code>SignatureDoesNotMatch</Code><Message>The request signature we calculated does not match the signature you provided. Check your key and signing method.</Message><StringToSignBytes>[snip]
```

A bit of investigation later, I found the problem only manifests when I've set `Content-Language` to `None`; setting it to another value, or not setting it at all, causes no problem.

The reason I'm setting it to None is because that's the value of `image.content_language`; I can obviously work around this in my code, but it seems like this shouldn't be the default behavior of boto.

I'm primarily testing with Python 2.6 and boto 2.19.0, but the issue still appears to be present in 2.24.0, the latest release.

zeckalpha:
Seems like this one was fixed along with #2477 in c8c625a. Does this still need to be open?

PR #2477, opened 2014-07-31
Add recognized HTTP headers

opottone:
I use S3 for static website hosting and want to add X-Robots-Tag HTTP header to some files. I was not able to  do this with boto, as it the header turned into x-amz-meta-x-robots-tag; not what I desired. However with this simple change to boto I was able to achieve what I wanted. While I was at it I did the same change for Content-Language header as well.

I don't completely understand when the x-amz-meta tag is necessary so there may be some catch, but at first glance this seems to work perfectly. Also it seems that this fixes bug https://github.com/boto/boto/issues/2050, although I'm not completely sure why. Then again, you might want to fix that bug for any metadata, not only for Content-Encoding.

### DIFF
--- a/boto/connection.py
+++ b/boto/connection.py
@@ -371,7 +371,7 @@ class HTTPRequest(object):
             for key in self.headers:
                 val = self.headers[key]
                 if isinstance(val, six.text_type):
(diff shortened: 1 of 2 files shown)


dnk8n:
This issue keeps coming up in a google search (similar problem but with boto3). Any direction on how to successfully achieve this with boto3?

PR #2557, opened 2014-08-26
Workaround for Python 2.7 httplib encoding bug

jlopex:
Python 2.7 httplib complains about paths generated by glacier tool.
This is patch forces ISO-8859-1 path encoding for all boto HTTP requests.


sloria:
+1. Working for me as well.

spg:
+1, worked for me as well

miccoli:
Same problem as https://github.com/boto/boto/pull/2697
both patches work for me.

PR #2578, opened 2014-08-29
Add support for PasswordResetRequired to IAM module

rvaralda:
Add support for PasswordResetRequired to create_login_profile and update_login_profile methods.

References:
http://docs.aws.amazon.com/IAM/latest/APIReference/API_CreateLoginProfile.html
http://docs.aws.amazon.com/IAM/latest/APIReference/API_UpdateLoginProfile.html


andrewalexander:
Is there a status on when this will get merged? I know that this is now available in boto3, but I would rather not have to migrate my code if I don't have to.

PR #2589, opened 2014-09-04
Miscellaneous improvements/fixes to cmdshell

alexras:
- Allow SSHClients to provide the number of retries to their connect()
  methods at construction time.
- Make the 'SSH Connection refused' print statement more informative.
- Since connection can fail during SSHClient construction without
  raising an exception, provide a method to ask whether we're really
  connected or not.
- Fix an error where `close()`ing an SSHClient created with
  `sshclient_from_instance` fails due to an undeclared method.


alexras:
Just checking up on this.

PR #2596, opened 2014-09-09
query_count fix in case of big size of item data

roman-yermishyn:
When table items size bigger than 1M, query_count returns wrong count. It returns count of items for the first data page only.
We need query all pages to receive correct items count.


pauloubuntu:
@danielgtaylor , @roman-yermishyn 

This issue has been solved by another pull request #2676 which has already been merged to develop branch.

You may close this PR.

Regards

PR #2606, opened 2014-09-12
SWF Layer2 region selection does not get passed to child objects

martinezah:
Support for specifying a region was recently added to SWF Layer2; however, child objects do not get passed the region configuration, and thus calls based on those objects (eg, child_obj.describe()) may fail or return invalid information.

This pull request passes the region to child objects as they are instantiated, ensuring region consistency across Layer2 objects.


laflechejonathan-amzn:
Any reason why this isn't merged in yet?

danielgtaylor:
The fix should now be in from #2985.

Issue #2549, opened 2014-08-22
Boto should support ELB Tags


adamhadani:
Elastic Load Balancer instances can have tags associated, and the AWS CLI tools allow retrieving them, e.g via the DescribeTags call. would be nice to have a instance.tags property on loadbalancer instance, similar to a EC2 tags on an ec2 instance.

reference:
http://docs.aws.amazon.com/cli/latest/reference/elb/describe-tags.html

rohit01:
+1

jamiemccrindle:
+1

s33dunda:
+1

cliff-wakefield:
Hmmm awscli has a means to perform the following on an ELB
- add tags (at create)
- add tags (retrospectively)
- remove-tags
- describe-tags

Features came in 1.4.1 which was released on 12th August 2014.

I thought awscli used boto exclusively?

Struggling to understand the awscli code base to see how they are doing it...

ryno75:
Cliff, I thought the same thing!  I asked a guy on the #boto irc channel and it turns out awscli actually uses botocore.  I am very tempted to fork boto, write the necessary methods to support get/set tags on ELB, and submit a merge request.  If I can find a bit of free time I probably will.

adamhadani:
@ryno75 I was very tempted to do the same myself but honestly found the
boto code very opaque and not friendly for outside developers to try and
fork/commit code to.. Hope you have better luck than me!

mfulleratlassian:
+1
Getting describe-tags functionality in would be a great start here

matteck:
+1
Need to programmatically set tags

jantman:
+1, because I'm cool like everyone else

shtuffit:
+1

ItsMadPig:
+1

codingfrenzy:
+1 for the ability to set tags to ELB programmatically

eleniS:
+1

oinume:
+1

tom-dalton-fanduel:
+1 just getting the commit above for describe tags merged would be a great start.

jamiegwatkin:
+1 really want this

smiller171:
:+1:

greenscar:
fd8063bda16e04fb4d113ecad54b7a951a89bb4b may not work for those trying to set variables, but it works for those trying to get variables. No reason to hold back getters.

shtuffit:
[...]

PR #2613, opened 2014-09-15
Add DescribeTags support for ELB instances (fixes: #2549)

Grindizer:
This patch add a "get_tags" method to Elastic Load Balancer objects.
It allows LB tags to be retrieved as a LbTagSet object (also dict)


accessomat:
+1 This is essential !

PR #2629, opened 2014-09-18
Fixed hive-site.xml installation while using InstallHiveStep

ghost:
Without --install-hive-site the hive-site.xml is not actually replaced.


achiku:
+1 for this p-r. Why it's not merged??

PR #2650, opened 2014-09-30
Prevent exceptions from being masked in get_contents_to_filename

llllllllll:
If opening the file fails, there is the potential for the following call to `os.remove` to fail, suppressing the actual error.


llllllllll:
ping, just wondering what the status of this issue is?

llllllllll:
ping

PR #2666, opened 2014-10-09
Updated to support api version 2013-11-15

mygu:
Using the QualificationRequirement Data Structure for Comparing Multiple Values (LocaleValue).

eg:
quals = Qualifications()
quals.add(LocaleRequirement("EqualTo", 'US', True))
quals.add(LocaleRequirement("In", ['US', 'CN', 'MX'], True))

### DIFF
--- a/boto/mturk/connection.py
+++ b/boto/mturk/connection.py
@@ -39,7 +39,7 @@ class MTurkRequestError(EC2ResponseError):
 
 class MTurkConnection(AWSQueryConnection):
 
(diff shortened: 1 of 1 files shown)


awatts:
Can this be updated for API version 2014-08-15? I'd like to be able to use all the new comparators and  LocaleValues with both Country and Subdivision (aka [US State Qualifications](http://mechanicalturk.typepad.com/blog/2015/01/us-state-qualifications.html))

PR #2667, opened 2014-10-09
Add support some data types in DynamoDB

kain-jy:
Add support `List`, `Map`, `Null`, `Boolean` data type.

### DIFF
--- a/boto/dynamodb/layer2.py
+++ b/boto/dynamodb/layer2.py
@@ -26,7 +26,7 @@ from boto.dynamodb.schema import Schema
 from boto.dynamodb.item import Item
 from boto.dynamodb.batch import BatchList, BatchWriteList
 from boto.dynamodb.types import get_dynamodb_type, Dynamizer, \
-        LossyFloatDynamizer
+        LossyFloatDynamizer, NonBooleanDynamizer
 
 
 class TableGenerator(object):
@@ -154,7 +154,7 @@ class Layer2(object):
                              profile_name=profile_name)
         self.dynamizer = dynamizer()
 
-    def use_decimals(self):
+    def use_decimals(self, use_boolean=False):
         """
         Use the ``decimal.Decimal`` type for encoding/decoding numeric types.
 
@@ -164,7 +164,7 @@ class Layer2(object):
 
         """
         # Eventually this should be made the default dynamizer.
-        self.dynamizer = Dynamizer()
+        self.dynamizer = Dynamizer() if use_boolean else NonBooleanDynamizer()
 
     def dynamize_attribute_updates(self, pending_updates):
         """
--- a/boto/dynamodb/types.py
+++ b/boto/dynamodb/types.py
@@ -52,7 +52,7 @@ def float_to_decimal(f):
 
 
 def is_num(n):
-    types = (int, long_type, float, bool, Decimal)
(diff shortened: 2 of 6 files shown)


xethorn:
@danielgtaylor : any update about this pull request / feature?

danielgtaylor:
Thanks for updating the code. I've taken another look and think this is a decent solution that won't break existing users. Thanks for contributing! :+1:

xethorn:
:+1: awesome

mattklein:
This doesn't seem to be working for me.

With this simple code:

```python
import boto
from boto.dynamodb.schema import Schema

conn = boto.dynamodb.connect_to_region(
    region_name='us-east-1',
    aws_access_key_id=ACCESS_KEY_ID,
    aws_secret_access_key=SECRET_ACCESS_KEY)
table = conn.table_from_schema(
    name='profile_stats',
    schema=Schema.create(hash_key=('profile_id', 'N')))

item1 = table.new_item(
    hash_key=98,
    attrs={
        'first': 32,
        'second': 45,
    }
)
item1.put()

item2 = table.new_item(
    hash_key=99,
    attrs={
        'first': 32,
        'second': 45,
        'the_map': {
            'first': 32,
            'second': 45,
        }
    }
)
item2.put()
```

item1 gets saved just fine, but item2.put fails with this error:

`
DynamoDBValidationError: DynamoDBValidationError: 400 Bad Request
{u'message': u'Supplied AttributeValue is empty, must contain exactly one of the supported datatypes', u'__type': u'com.amazon.coral.validate#ValidationException'}
`

I get the same results with each of boto-2.35.2, boto-2.36.0, boto-2.37.0, and boto-2.38.0.

If I'm doing something wrong, it sure isn't obvious to me! Can anybody shed some light?

mattklein:
After some more debugging, it looks like the class boto.dynamodb.layer1.Layer1 is coded to hit the original (20111205) version of the DynamoDB HTTP API ([old apidocs](http://docs.aws.amazon.com/amazondynamodb/latest/developerguide/API_PutItem_v20111205.html)). But that version of the API doesn't support the Map data type -- only the newer version (20120810) does ([new apidocs](http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_PutItem.html)).

This is from the Layer1 class:

```python
class Layer1(AWSAuthConnection):
    """
    This is the lowest-level interface to DynamoDB.  Methods at this
    layer map directly to API requests and parameters to the...
    """

    DefaultRegionName = 'us-east-1'
    """The default region name for DynamoDB API."""

    ServiceName = 'DynamoDB'
    """The name of the Service"""

    Version = '20111205'
    """DynamoDB API version."""

```

I tried simply switching the Version property to use 20120810 rather than 20111205, but then the library constructs requests that aren't in conformance with the API's schema:

```
DynamoDBValidationError: DynamoDBValidationError: 400 Bad Request
{u'message': u'The provided key element does not match the schema', u'__type': u'com.amazon.coral.validate#ValidationException'}
```

So obviously it wouldn't be as simple a change as this to upgrade to the new version of the API.

But I'm still fundamentally confused -- it sure seems to me that Boto doesn't currently support the Map data type, despite what this patch intends. How is this feature working for anybody? Is it maybe just the .get() but not the .put() that works?

PR #2678, opened 2014-10-14
Modify vpc attr

vasileiosm:
ModifySubnetAttribute implementation under boto/ec2.

As @markdoliner pointed out, this action was added in a newer version of AWS API thus I added the option to override the Version parameter explicitly.

### DIFF
--- a/boto/connection.py
+++ b/boto/connection.py
@@ -1107,7 +1107,7 @@ class AWSQueryConnection(AWSAuthConnection):
                                                     self.host)
         if action:
             http_request.params['Action'] = action
-        if self.APIVersion and http_request.params.get('Version') is None:
+        if self.APIVersion and http_request.params.get('Version', None) is None:
             http_request.params['Version'] = self.APIVersion
         return self._mexe(http_request)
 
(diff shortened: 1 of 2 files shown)


gkze:
Any updates on this?

Fandekasp:
Please give an update on this ticket, thank you!

FlorinAndrei:
Very useful feature, can't imagine why it hasn't been merged yet.

geradcoles:
Any chance of this making it in? I have a fully scripted deployment that creates a with subnets, auto-scaling groups, etc, with an ElastiCache Redis cluster. We need remote access to the instances in the ElastiCache cluster, but since there is no api to create public subnets, my elasticache cluster is only available on private addresses from within the VPC. Our other instances are build with launch configuration, where we can enable public IP's, but this part is a major flaw since there isn't any good way to get remote access to these Redis systems without deploying NAT instances or similar hacky solutions.

radekg:
Wow, this still isn't in? Nearly a year...

fermayo:
Incredible!

amegianeg:
:+1:

JordonPhillips:
We will need tests (and a rebase) before we can consider this.

PR #2694, opened 2014-10-20
[MWS] Remove unnecessary debug print statement

ckrybus:


disruptek:
I decided this doesn't need tests in my branch.  Thanks for the catch.

Issue #2657, opened 2014-10-04
CloudSearch2 not working with IAM-based search requests


ndp805:
Per http://aws.amazon.com/about-aws/whats-new/2014/08/14/amazon-cloudsearch-adds-enhanced-iam-integration/ CloudSearch2 (boto 2.32.1) does not appear to be properly signing/submitting search requests against domain which has authorized specific EC2 IAM roles

Running the appropriate python snippet on an EC2 instance (Amazon LINUX) with an IAM Role with even administrative privileges:

```
    try:
        conn = Layer2(region=region)
        domain = conn.lookup(domain_name='....')
        search_service = domain.get_search_service()
        query_results = search_service.search(q='.....', start=0, size=20, parser='simple')
        ....
    except Exception as e:
        traceback.print_exc()
        ....
```

Yields:

'Unknown error processing search {"message": "User: anonymous is not authorized to perform: cloudsearch:search on resource: arn:aws:cloudsearch:us-west-2:{account}:domain/{name}", "__type": "#AccessDenied",.....}...."

Have even tried using IAM policy on the search domain itself:

{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "",
      "Effect": "Allow",
      "Principal": {
        "AWS": "arn:aws:iam::<account-id>:role/{role-name}"
      },
      "Action": [
        "cloudsearch:suggest",
        "cloudsearch:search"
      ]
    }
  ]
}

Same issue.

Have subsequently run the new AWS CLI on the same host (aws-cli/1.4.4 Python/2.6.9 Linux/3.14.19-17.43.amzn1.x86_64) and this _does_ appear to work properly:

% aws cloudsearchdomain search --search-query='hello' --endpoint-url https://{search-domain-endpoint}

....( Results as expected ) ...

Believe the latest CLI cloudsearchdomain enhancement leverages new boto core which properly signs the request (?)

PR #2717, opened 2014-10-29
Fix CloudSearch2 to work with IAM-based search and upload requests

andersenyang:
Fixes https://github.com/boto/boto/issues/2657.
Correctly signs headers when submitting batch uploads and search requests.


skunkwerk:
+1 for a new release with this (much appreciated)

akshayjshah:
Likewise!

adepue:
Thanks for the new release! 
Blocking issue in it with dynamo though that is addressed here:
https://github.com/boto/boto/pull/2871

PR #2718, opened 2014-10-29
Fix str encode for request with proxy (Py3)

drgarcia1986:
to fix this error:

``` python
  File "<pyshell#6>", line 1, in <module>
    conn.send_email('foo@bar.com', 'Foo!', 'Bar', ['bar@foo.com])
  File "C:\Python34\lib\site-packages\boto\ses\connection.py", line 276, in send_email
    return self._make_request('SendEmail', params)
  File "C:\Python34\lib\site-packages\boto\ses\connection.py", line 102, in _make_request
    data=urllib.parse.urlencode(params)
  File "C:\Python34\lib\site-packages\boto\connection.py", line 1068, in make_request
    retry_handler=retry_handler)
  File "C:\Python34\lib\site-packages\boto\connection.py", line 913, in _mexe
    self.is_secure)
  File "C:\Python34\lib\site-packages\boto\connection.py", line 705, in get_http_connection
    return self.new_http_connection(host, port, is_secure)
  File "C:\Python34\lib\site-packages\boto\connection.py", line 747, in new_http_connection
    connection = self.proxy_ssl(host, is_secure and 443 or 80)
  File "C:\Python34\lib\site-packages\boto\connection.py", line 796, in proxy_ssl
    sock.sendall("CONNECT %s HTTP/1.0\r\n" % host)
TypeError: 'str' does not support the buffer interface
```

### DIFF
--- a/boto/connection.py
+++ b/boto/connection.py
@@ -793,18 +793,18 @@ class AWSAuthConnection(object):
         else:
             sock = socket.create_connection((self.proxy, int(self.proxy_port)))
         boto.log.debug("Proxy connection: CONNECT %s HTTP/1.0\r\n", host)
-        sock.sendall("CONNECT %s HTTP/1.0\r\n" % host)
-        sock.sendall("User-Agent: %s\r\n" % UserAgent)
+        sock.sendall(("CONNECT %s HTTP/1.0\r\n" % host).encode())
+        sock.sendall(("User-Agent: %s\r\n" % UserAgent).encode())
         if self.proxy_user and self.proxy_pass:
             for k, v in self.get_proxy_auth_header().items():
-                sock.sendall("%s: %s\r\n" % (k, v))
+                sock.sendall(("%s: %s\r\n" % (k, v)).encode())
             # See discussion about this config option at
             # https://groups.google.com/forum/?fromgroups#!topic/boto-dev/teenFvOq2Cc
             if config.getbool('Boto', 'send_crlf_after_proxy_auth_headers', False):
-                sock.sendall("\r\n")
+                sock.sendall(("\r\n").encode())
         else:
-            sock.sendall("\r\n")
-        resp = http_client.HTTPResponse(sock, strict=True, debuglevel=self.debug)
+            sock.sendall(("\r\n").encode())
+        resp = http_client.HTTPResponse(sock, debuglevel=self.debug)
         resp.begin()
 
         if resp.status != 200:


rjpereira:
Not sure if I understand the current status: this fix has not yet been merged, right ?

felix-d:
@rjpereira it seems not, I still get this error -.-

Rusty-1:
Is there a viable work around using boto

citosid:
Is there any solution to this?

felix-d:
@citosid you can use the branch. I forked boto myself and applied that code change and never had issues since.

citosid:
Thank you @felix-d. I've applied the same fix in the server I have the code... but I just wish to use the main project, why the merge request was never accepted?

felix-d:
@citosid I have no idea ^^ it's pretty old... maybe other solutions exist today

pfriel-bcs:
:+1: Would like to see this merged, I ran into this problem today while updating all of our scripts to Python 3.x.

lawrencexia:
:+1: Running into this issue and would like to see this merged!

jrandall:
This PR fixes a critical issue with Python 3 support in the S3 service, yet it has gone 30 months without being merged. @jamesls you tagged it as "needs-tests" which would be a nice thing to have, but in this case the code change is trivial and easy to verify as correct, while the test is relatively hard to do properly because you really need to have a real connection to a S3 service exercise in order to test it properly, and I don't see any existing tests that work that way. I suppose we could mock the `sock.sendall` method, but it will not be trivial to make a useful test of these changes.

colin-nolan:
Given that this PR is tagged with "needs-tests", I have resolved this here: https://github.com/boto/boto/pull/3699. Please may we now finally get this issue resolved?

xiaket:
I also would like to see this merged. I ran into this issue this morning using ansible and it takes me a solid hour to trace it down.

xiaket:
I have found an inappropriate behaviour in the original commit(8ee04fd). The `strict=True` is removed from the HTTPResponse instance initialization, but this can lead to unexpected behaviour for python2 users.

This keyword argument in python2 essentially allowed incoming HTTP/0.9 traffic, which should be discouraged as a security measure. The original source code correctly handled this by setting `strict` to `True` to disallow weird response be recognised as HTTP/0.9 traffic.

The behaviour of this keyword argument is changed in python3(pre 3.4), which gives out a DeprecationWarning regardless the provided value of the keyword argument. And finally, this keyword argument is removed in python3.4, and if you provide `strict` as a keyword argument, a TypeError will be raised.

As a conclusion, we should enable this keyword argument for python2 and disable it for python3, which leads to the following code:

```
 -        resp = http_client.HTTPResponse(sock, strict=True, debuglevel=self.debug)
 +        keywords = {'sock': sock, 'debuglevel': self.debug}
 +        if six.PY2:
 +            keywords['strict'] = True
 +        resp = http_client.HTTPResponse(**keywords)
```

However, this extra fix might not be necessary since all we going to parse are from AWS...

BTW, can some who have write access to this repo approve this pull request please? This issue will affect all python3 users behind a proxy, and they cannot do anything with AWS. Since boto is widely used as a base library for everything involving AWS, this can affect more people too.

colin-nolan:
@xiaket Perhaps you could make that change and pull request to [wtsi-hgi:fix/3561-py3-proxy-ssl-connection](https://github.com/wtsi-hgi/boto/tree/fix/3561-py3-proxy-ssl-connection), where I've already extended this work with tests?

Ideally the author of this request would consolidate all of the changes but they don't appear to be active here (not surprising given the age of this PR!).

xiaket:
Thanks @colin-nolan ! I've done that, and hopefully, someone will merge this soon.

colin-nolan:
Thanks @xiaket - I've merged your work into https://github.com/boto/boto/pull/3699. One day someone might merge that pull request (@jamesls?)...

chris-goffinet-ck:
Can this get merged? It still isn't merged.. Its critical to be able to use S3 + Proxy with boto.

ps-jay:
Also needed for Route53 + proxy

johnboy2:
Also needed for IGW + proxy.

PR #2758, opened 2014-11-13
fix keypair.save() to work with Python3 (unicode/byte strings)

hjacobs:
Quickfix to make KeyPair.save() work with Python 3 (self.material is a str in Python 2 and Python 3, so the file needs to be opened in text mode).

Should work with Python 2.6 (uses "with" statement) and up.


hjacobs:
Outdated.

PR #2804, opened 2014-12-10
Allow ignoring of the default `endpoints.json`

goakley:
Provide a config value that allows boto to ignore the default
`endpoints.json` file when loading region info.

This is in response to an issue where zipped python executables are
unable to access the default file, since it is contained in an
archive and cannot be accessed with `open()`.


goakley:
Fixed with the use of the BOTO_ENVIRONMENT environment variable.

PR #2809, opened 2014-12-15
DynamoDB table batch_get fails to process all remaining results if single batch result is empty.

garthberry:
If the batch_get keys contain a contiguous block of keys that do not exist in the target table that is greater than or equal to max_batch_get then the remaining results will no be processed even if there are still keys waiting.

I have added an integration test to simulate this condition. The eventually consistent batch test has been updated to contain two keys that do not exist in the target table. The max_batch_get property has been set to 2. This reliably produces the failure described above, no items are returned by the batch_get instead of the 2 expected.


adepue:
This commit has caused infinite looping when no results should be returned for us.

PR #2812, opened 2014-12-16
Add query string to body for anon STS POST

kyleknap:
So this PR will solely affect anonymous requests using STS. Before, the request would put the query string into the URL path. However, all requests for STS are modeled as a post and that data should be in the body. This PR removes the query string from the URL path and adds it to the body. I tested it out using some of the different ways to assume a role and it looks like it is working fine.

cc @danielgtaylor @jamesls

### DIFF
--- a/boto/auth.py
+++ b/boto/auth.py
@@ -771,16 +771,15 @@ class S3HmacAuthV4Handler(HmacAuthV4Handler, AuthHandler):
                                     urllib.parse.urlencode(req.params))
 
 
-class QueryAuthHandler(AuthHandler):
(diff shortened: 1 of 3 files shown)


aclaypool:
@kyleknap Sorry to "bring up old s***" but we used the anon STSConnection facilitate CLI authentication using a SAML response from Active Directory Federation Services in 2.34.0...I think this upgrade appears to have broken our requests now when we are testing 2.38.0.  We are getting nothing but 400 errors when we try to assume_role_with_saml from an Anon STS connection.  Is there a change I need to make for the STSConnection other than anon=True?

PR #2815, opened 2014-12-16
Add support for the 'cursor' parameter in cloudsearch

felipetrz:
Adds support for the cursor parameter on cloudsearch, which is missing, and also makes the next_page() method of the result use the cursor when it's available, allowing us to iterate through the whole result set, unrestrained by amazon's 10000 limit.

### DIFF
--- a/boto/cloudsearch2/search.py
+++ b/boto/cloudsearch2/search.py
@@ -83,6 +83,9 @@ class Query(object):
                  facet=None, highlight=None, partial=None, options=None,
                  cursor=None):
 
+        if cursor is not None:
+            start = None  # start can't coexist with cursor
+
         self.q = q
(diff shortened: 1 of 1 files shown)


natecode:
There are a few bugs with this patch. First, it doesn't add cursor support to `SearchConnection.to_domain_connection_params()`. For that you need to add to cloudsearch2/search.py:

@@ -199,6 +217,9 @@
         if self.sort:
             params['sort'] = ','.join(self.sort)

+        if self.cursor:
+            params['cursor'] = self.cursor
+
         return params

Also, it does not update `SearchConnection.get_all_hits()` and `get_all_paged()` to work with cursors. I'm not sure how to implement that change consistently there. Perhaps one of you can?

PR #2819, opened 2014-12-17
Call logger.exception correctly

graingert:
https://docs.python.org/2/library/logging.html#logging.Logger.exception "This method should only be called from an exception handler."


danielgtaylor:
@graingert would it be possible to relicense this PR as MIT? That's Boto's license, which is preferred.

graingert:
Of course this PR is licenced under MIT please note from your CONTRIBUTING file:

> Must be appropriately licensed (New BSD).

graingert:
This PR is licensed (MIT).

PR #2823, opened 2014-12-17
Fix BOTH_PATH to work with Windows drives

thobrla:
This works around the issue that Windows paths containing drive letters have the ":" character, causing BOTO_PATHs including Windows drives to split incorrectly.


thobrla:
Ping - anything else you would like here?

Issue #2828, opened 2014-12-19
Unable to call query_2 after call describe method - DynamoDB2 module


pauloubuntu:
Hi,

I've noticed a weird behaviour when trying to query anything after called describe method. Here is an example:

Script that fails because of describe method is called before query:
```Python
from boto.dynamodb2.table import Table

invitationTable = Table("INVITATION")
invitationTable.describe()

oldInvitationResultSet = invitationTable.query_2(
    index='date_when_gsi-index',
    date_when_gsi__eq='20141226',
    max_page_size=20
)

for user in oldInvitationResultSet:
    print user['owner_user_name'], user['user_name']
```

That outputs :
```
Traceback (most recent call last):
  File "/Users/paulo.rodenas/workspaceIdea/brejaadmin/02-Database/02-Routine/02-InvitationsRemoval/OldInvitationRemoval/script.py", line 15, in <module>
    max_page_size=20
  File "/Users/paulo.rodenas/workspaceIdea/boto/boto/dynamodb2/table.py", line 984, in query_2
    "You must specify more than one key to filter on."
boto.dynamodb2.exceptions.QueryError: You must specify more than one key to filter on.
```

The same script but without calling describe method:

```Python
from boto.dynamodb2.table import Table

invitationTable = Table("INVITATION")
#invitationTable.describe()

oldInvitationResultSet = invitationTable.query_2(
    index='date_when_gsi-index',
    date_when_gsi__eq='20141226',
    max_page_size=20
)

for user in oldInvitationResultSet:
    print user['owner_user_name'], user['user_name']
```

That outputs:

```
GUIJOU CHARLOUMICH
GUIJOU HDRI
GUIJOU GODASSE45

Process finished with exit code 0
```

pauloubuntu:
Closing this issue since the PR #2829 has been merged.

xethorn:
:+1:

Issue #2708, opened 2014-10-26
boto: Missing global_indexes, boto.dynamodb2.exceptions.QueryError: You must specify more than one key to filter on


xethorn:
This exception happens randomly.

Scenario:
- For my unit tests, I have a `create_user` in my fixtures that creates a user (values are all the same, just a number is being added.)
- Create user fetches after that specific user to verify it was created.
- All fetches are working until a random one.
- query_2 is called with: 

```
limit = 1
index = 'username_index'
reverse = False
consistent = False
attributes = None
max_page_size = None
query_filter = None
conditional_operator = None
filter_kwargs = {'username__eq': 'test5'}
```
- `self.schema`:
```
(Pdb) self.schema
[<boto.dynamodb2.fields.HashKey object at 0x3633fd64cc0>]
```

- `self.global_indexes` is empty.

Schema of the table:
```
(Pdb) self.describe()
{'Table': {
    'ProvisionedThroughput': {'WriteCapacityUnits': 1, 'NumberOfDecreasesToday': 0, 'ReadCapacityUnits': 1}, 
    'AttributeDefinitions': [
        {'AttributeName': 'email', 'AttributeType': 'S'}, 
        {'AttributeName': 'id', 'AttributeType': 'S'}, 
        {'AttributeName': 'username', 'AttributeType': 'S'}], 
    'ItemCount': 10, 
    'GlobalSecondaryIndexes': [
        {
            'IndexStatus': 'ACTIVE', 
            'IndexSizeBytes': 1785, 
            'Projection': {'ProjectionType': 'ALL'}, 
            'ItemCount': 10, 
            'ProvisionedThroughput': {'WriteCapacityUnits': 1, 'NumberOfDecreasesToday': 0, 'ReadCapacityUnits': 1}, 
            'KeySchema': [
                {'KeyType': 'HASH', 'AttributeName': 'username'}], 
            'IndexName': 'username_index'
        }, {
            'IndexStatus': 'ACTIVE', 
            'IndexSizeBytes': 1785, 
            'Projection': {'ProjectionType': 'ALL'}, 
            'ItemCount': 10, 
            'ProvisionedThroughput': {
                'WriteCapacityUnits': 1, 
                'NumberOfDecreasesToday': 0, 
                'ReadCapacityUnits': 1}, 
            'KeySchema': [
                {'KeyType': 'HASH', 'AttributeName': 'email'}], 
            'IndexName': 'email_index'
        }], 
    'TableName': 'user', 
[...]

PR #2829, opened 2014-12-19
Fixed bug that unable call query_2 after call describe method on dynamodb2 module

pauloubuntu:
This fixes the #2828 and #2708 issues.

As shown on that issue, there was a strange behaviour when calling describe method and then query_2. 

I've found out that when you create a fresh new Table instance it has not schema, indexes or even global_indexes. So when you call query_2 method is check whether you have those parameters and if you haven't it tries to perform the call as usual. 

In the other hand when you call describe method it was setting all those instance variables but self.global_indexes which was causing it to raising an exception on the following statement at the begging of query_2 method as shown below;

```Python
 if self.schema:
            if len(self.schema) == 1:
                if len(filter_kwargs) <= 1:
                    if not self.global_indexes or not len(self.global_indexes):
                        # If the schema only has one field, there's <= 1 filter
                        # param & no Global Secondary Indexes, this is user
                        # error. Bail early.
                        raise exceptions.QueryError(
                            "You must specify more than one key to filter on."
                        )
```

As Global Secondary Indexes were being sent by the DynamoDB API, all I had to do was to update the instance variable self.global_indexes with that information.  Now it's working as expected :)

I hope you find my pull request useful. If you do, please accept my pull request.

Best Regards

### DIFF
--- a/boto/dynamodb2/table.py
+++ b/boto/dynamodb2/table.py
@@ -297,6 +297,39 @@ class Table(object):
 
         return indexes
 
+    def _introspect_global_indexes(self, raw_global_indexes):
+        """
+        Given a raw global index structure back from a DynamoDB response, parse
+        out & build the high-level Python objects that represent them.
+        """
+        indexes = []
+
+        for field in raw_global_indexes:
+            index_klass = GlobalAllIndex
+            kwargs = {
+                'parts': []
+            }
+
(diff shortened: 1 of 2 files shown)


xethorn:
Ping @danielgtaylor?

danielgtaylor:
LGTM, thanks! :+1:

PR #2831, opened 2014-12-19
Allow additional query parameters to be passed to generate url

chrisseto:
For the ability to create pre signed url for versions, alc, etc.

### DIFF
--- a/boto/s3/bucket.py
+++ b/boto/s3/bucket.py
@@ -624,10 +624,11 @@ class Bucket(object):
 
     def generate_url(self, expires_in, method='GET', headers=None,
                      force_http=False, response_headers=None,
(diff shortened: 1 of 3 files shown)


danielgtaylor:
@chrisseto can you give an example of when this would be used? What can you not do today without adding this?

Also, for it to be merged you'll need to add some tests covering the new feature.

chrisseto:
@danielgtaylor In an application I am helping develop we use boto to create signed URLs but there is no way to get a listing of versions via signed URL.
Adding in the option to tack on additional query parameters will include them in the signature resulting in a valid URL. 
This is the most flexible to add these options in but their purpose is not very explicit.

Issue #2681, opened 2014-10-15
Does boto.rds2 work on Python3 . Is there a way to access the boto.rds2  method in python


henin:
I was trying to get all db_instances using rds2. Using res.get_all_dbinstances() i was able to get all db_instances but in rds2 i am getting the below error
import boto.rds2
import json

con=boto.rds2.connect_to_region("ap-southeast-1")
raw=con.describe_db_instances(db_instance_identifier=None)

Gives you an error : File "/usr/local/Cellar/python3/3.4.1_1/Frameworks/Python.framework/Versions/3.4/lib/python3.4/json/init.py", line 312, in loads
s.class.name))
TypeError: the JSON object must be str, not 'bytes'

Bustead:
@henin Good news! #2834 resolves this issue. Do you mind closing this issue? Happy days! :)

Issue #2677, opened 2014-10-14
TypeError: the JSON object must be str, not 'bytes' when using boto.rds2.describe_db_instances() using python 3.4.1


henin:
want to get all the running instances in a AWS environment. I am using python 3.3. i had used boto.rds earlier which has a method called rds2.get_all_dbinstances(). But i am migrating now to rds2. when i try to run the above code using python interpreter usage :python2 filename.py : i get no error. the program works fine. But using Bash shell usage : ./filename.py : i get the error :

TypeError: File "/Users/heninkarkada/Documents/repositories/jaws/b.py", line 5, in <module>
    raw=con.describe_db_instances(db_instance_identifier=None)
  File "/usr/local/lib/python3.4/site-packages/boto/rds2/layer1.py", line 1512, in describe_db_instances
    path='/', params=params)
  File "/usr/local/lib/python3.4/site-packages/boto/rds2/layer1.py", line 3764, in _make_request
    return json.loads(body)
  File "/usr/local/Cellar/python3/3.4.1_1/Frameworks/Python.framework/Versions/3.4/lib/python3.4/json/__init__.py", line 312, in loads
    s.**class**.**name**))
TypeError: the JSON object must be str, not 'bytes'the JSON object must be str, not 'bytes' at line no:3
Can anyone tell me why this thing is happening? And is RDS2 compatible with python3 , so can you  please fix this defect  Advance thanks

swimmadude66:
This issue is still happening, and can be fixed by simply adding:

`body = response.read().decode('utf-8')` 

at line 3674 of rds2/layer1.py

rbruggem:
Any updates on this?

stefanhorning:
Still necessary and swimmadude66's fix still works, line has moved to `3774` in the meantime. Surprised this still wasn't fixed!

PR #2834, opened 2014-12-21
Fixes JSON error parsing error when receiving bytes in Python 3.

codebyneil:
Should fix a couple errors:
https://github.com/boto/boto/issues/2681
https://github.com/boto/boto/issues/2677

That said, I had a bit of trouble with testing, so I'm not 100% on this... If you're going to try to test it, I suggest removing the try/except wrapping because I was getting a different TypeError in the testing framework than I was from the actual bug.

### DIFF
--- a/boto/rds2/layer1.py
+++ b/boto/rds2/layer1.py
@@ -3760,6 +3760,8 @@ class RDSConnection(AWSQueryConnection):
                                      path='/', params=params)
         body = response.read()
         boto.log.debug(body)
+        if isinstance(body, bytes):
+            body = body.decode('utf-8')
         if response.status == 200:
             return json.loads(body)
         else:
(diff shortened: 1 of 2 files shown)


tuxlife:
Hello @danielgtaylor,

is something missing, or can you please merge the pr?

many thanks,
Matthias

StevenMapes:
Any word on when this is likely to be merged in as it's been almost a year and it fixes a very real bug

mochawich:
will this be merged anytime soon❓
Thanks

the-neil:
I don't even remember writing this anymore... 😛 Have you guys still been
seeing this error?

mochawich:
Yes, using [Ansible - Rds module](http://docs.ansible.com/ansible/latest/rds_module.html) with Python 3 results in this error.

baughj:
@nourchawich 👍 to that. Any reason why this cannot be merged?

PR #2842, opened 2014-12-28
DynamoDB2.table.Table: Use of `connection` kwargs

### LINKED ISSUE #1585: Unable to (easily) specify region for dynamodb2
According to the docs for 2.9.6, you should be able to specify the region on the Table constructor as:

```
from boto.dynamodb2.layer1 import DynamoDBConnection
users = Table(table_name='users', 
                  schema=[HashKey('username')],
                  connection=DynamoDBConnection(
                      region='us-west-2'
                  ))
```
 
Unfortunately, that throws an error because the ```__init__``` for the DynamoDBConnection assumes the ```region``` is actually a ```RegionInfo``` object not a string and calls ```region.endpoint``` on line https://github.com/boto/boto/blob/develop/boto/dynamodb2/layer1.py#L70 and throws an error.

```
  File "/tmp/test.py", line 54, in _get_table
    connection=DynamoDBConnection(region='us-west-2'))
  File "/tmp/.virtualenvs/test/lib/python2.7/site-packages/boto/dynamodb2/layer1.py", line 70, in __init__
    kwargs['host'] = region.endpoint
AttributeError: 'str' object has no attribute 'endpoint'
```
If ```region``` is not specified, it fetches the region name from the boto config file.

It looks like the only way to call it with a different region from us-east-1 right now is to use the connect_to_region method in boto.dynamodb2:

```
from boto.dynamodb2.layer1 import DynamoDBConnection
import boto
users = Table(table_name='users',
                  schema=[HashKey('username')],
                  connection=boto.dynamodb2.connect_to_region('us-west-2')))
```

Is this the correct way to do it?  If so - I can submit a PR with an update to the docs.

quentindemetz:
fixes #1585
Clarify the existing comment to show how to manipulate tables in another region.

### DIFF
--- a/docs/source/dynamodb2_tut.rst
+++ b/docs/source/dynamodb2_tut.rst
@@ -89,8 +89,8 @@ A full example::
     ...         HashKey('account_type', data_type=NUMBER),
     ...     ])
     ... ],
-    ... # If you need to specify custom parameters like keys or region info...
-    ... connection= boto.dynamodb2.connect_to_region('us-east-1'))
+    ... # If you need to specify custom parameters, such as credentials or region, use the following:
+    ... Table.create('users', connection=boto.dynamodb2.connect_to_region('us-east-1'))
 
 
 Using an Existing Table


danielgtaylor:
Thanks! :+1:

PR #2843, opened 2014-12-29
Expose cloudformation `UsePreviousTemplate` param

### LINKED ISSUE #2628: Support for parameters-only update with CloudFormation (adding UsePreviousTemplate kwarg to update_stack)
For some time now CloudFormation has been supporting updates without the need to provide the template every time:
http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_UpdateStack.html (search for ``UsePreviousTemplate``)
Adding this to boto should be quick and it will be a big win as it will reduce the number of calls services that use CloudFomration need to make.

cosmopetrich:
When provided, this parameter causes cloudformation to reuse the current
template, rather than a new one provided via TemplateBody or
TemplateURL.

Note that the CloudFormation API will explicitly reject requests containing
both `UsePreviousTemplate` and at least one of `TemplateBody` and
`TemplateURL` with the error `You cannot specify both
usePreviousTemplate and Template Body/Template URL`.

Please let me know your thoughts.

This commit fixes #2628

### DIFF
--- a/boto/cloudformation/connection.py
+++ b/boto/cloudformation/connection.py
@@ -98,7 +98,8 @@ class CloudFormationConnection(AWSQueryConnection):
     def _build_create_or_update_params(self, stack_name, template_body,
             template_url, parameters, disable_rollback, timeout_in_minutes,
             notification_arns, capabilities, on_failure, stack_policy_body,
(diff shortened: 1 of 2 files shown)


danielgtaylor:
This looks great, thanks! :+1:

PR #2844, opened 2014-12-30
Fix S3 sigv4 signature when non-ascii present in query string or tilde characters present in key name

pslawski:
This fixes incorrect sigv4 signature for S3 requests (#2846, #2849):

1. When non-ascii characters are present in request's query string in Python 2. For calculation of the signature, the query string is parsed and placed into the mangled request's parameters. This is done by parsing the query string using `urllib.parse.parse_qs`. The query string given is in Unicode which in Python 2 `parse_qs` doesn't do decoding of URL-encoded %HH bytes. When the query string has URL-encoded non-ascii characters, the wrong Unicode string is returned.
   * e.g. a value of u'%C3%B1' would be outputted from `parse_qs` as u'\xc3\xb1', but originally the value url encoded was b'\xc3\xb1' (utf-8 encoded bytes) or u'\xf1' in Unicode.

2. When tilde characters `~` are present in the key name since the Canonical URI URL-encodes `~` which is incorrect by http://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html

### DIFF
--- a/boto/utils.py
+++ b/boto/utils.py
@@ -104,6 +104,33 @@ def unquote_v(nv):
         return (nv[0], urllib.parse.unquote(nv[1]))
 
 
+def parse_qs_safe(qs, keep_blank_values=False, strict_parsing=False,
+                  encoding='utf-8', errors='replace'):
+    """Parse a query handling unicode arguments properly in Python 2."""
+    if six.PY3:
+        return urllib.parse.parse_qs(qs, keep_blank_values, strict_parsing,
+                                     encoding, errors)
+    else:
+        is_text_type = isinstance(qs, six.text_type)
+        if is_text_type:
+            qs = qs.encode('ascii')  # URL encoding uses ASCII code points only
+
+        qs_dict = urllib.parse.parse_qs(qs, keep_blank_values, strict_parsing)
+
+        if is_text_type:
+            return decode_qs_dict(qs_dict, encoding, errors)
+        return qs_dict
+
+
+def decode_qs_dict(qs_dict, encoding='utf-8', errors='replace'):
+    result = {}
+    for (name, value) in qs_dict.items():
+        decoded_name = name.decode(encoding, errors)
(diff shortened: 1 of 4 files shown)


Captainkrtek:
+1

tpodowd:
+1 on this also. I tried it out and it works well although I didn't run the tests but I tested on the command line and it did fix a problem regarding bad signature when using a Japanese prefix.

kouk:
:+1:

jamesls:
Taking a look.

boto/utils.py
```diff
@@ -104,6 +104,33 @@ def unquote_v(nv):
         return (nv[0], urllib.parse.unquote(nv[1]))
 
 
+def parse_qs_safe(qs, keep_blank_values=False, strict_parsing=False,
+                  encoding='utf-8', errors='replace'):
```
jamesls: What's the motivation for having `errors='replace'`?  Is there a case where a `.decode()` in `decode_qs_dict` would fail to decode?  If I change this to `strict`, the tests you've added still pass.

boto/utils.py
```diff
@@ -104,6 +104,33 @@ def unquote_v(nv):
...
+    """Parse a query handling unicode arguments properly in Python 2."""
+    if six.PY3:
```
jamesls: I'm going to go ahead and move this over to `compat.py`.

jamesls:
Seems reasonable.  I'm going to go ahead and make a few small structural changes and then merge.

boto/utils.py
jamesls: Ah, I see, compatibility with the default values for `parse_qs`.  Seems reasonable.

jamesls:
Merged via 5092c6d.  Thanks for the pull request.

tpodowd:
Thanks @jamesls !  Good to get this stuff in.

PR #2850, opened 2014-12-30
Set version_id to None when 'null' is returned from s3

ghost:
Currently, when versioning is turned on or suspended for an s3 bucket
calling `key.generate_url` will include a `versionId=null` when the key
in question returns 'null' from s3 as the version id. This will cause
a 403 when trying to use it.

### DIFF
--- a/boto/s3/key.py
+++ b/boto/s3/key.py
@@ -238,7 +238,8 @@ class Key(object):
         # version of the Key.  In that case, we don't really want to
         # overwrite the version_id in this Key object.  Comprende?
         if self.version_id is None or force:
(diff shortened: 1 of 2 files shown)


kivinju:
I encounter the same issue. I will have a versionId=null in my url.

esmail:
This issue was recently causing me problems with `S3BotoStorage` from [`django-storages`](https://github.com/jschneier/django-storages). 

The proposed fix didn't help, as I was getting `resp.getheader(provider.version_id, None) == None`, but upgrading to `boto3`/`S3Boto3Storage` worked.

PR #2857, opened 2015-01-05
Fix #2836

### LINKED ISSUE #2836: Can't use bucket names with dots
Using boto for a s3 bucket named with dots, e.g. my.bucket.s3.amazonaws.com fails:

ssl.CertificateError: hostname 'my.bucket.s3.amazonaws.com' doesn't match either of '*.s3.amazonaws.com', 's3.amazonaws.com'

krallin:
This patch proposes a fix for https://github.com/boto/boto/issues/2836

Specifically, it ensures that SSL hostname validation is indeed disabled when `Connection. https_validate_certificates` is `False`. 

When ``Connection.https_validate_certificates` is `True`, this patch ensures that SSL hostname validation is performed using boto's own validation function (`ValidateCertificateHostname`), across all versions of Python (and that no redundant validation is performed by the standard library).

Note that the failing test only fails on Python 2.7.9 (in that version, the standard library defaults to hostname validation).


krallin:
Ping — just wondering if there was anything I could do to help get this moving forward? 

(or to know whether there was something fundamentally wrong this approach)

Thanks!

rh0dium:
Ping!  Please consider this patch!

krallin:
Ping..?

JernejJerin:
I was having this problem when running Django collectstatic, when using buckets with dots in name. I applied your solution. Now collectstatic runs, but when executing I receive this: 
```
socket.error: [Errno 10054] An existing connection was forcibly closed by the remote host
```

dogweather:
:+1:

jheasly:
+1 Patch worked for me.

7heo:
@jheasly This patch works, but it is a non-solution that will lower the security standards.
@krallin It's great that you made a patch that works, but as wrote on the line above, it is a non-solution that will bring more problems in, since you basically disable SSL checks silently (MITM alert) when using dotted names. I strongly warn (since I can't oppose) against merging this patch. Also, I am working on a proper fix which wouldn't hinder security.

krallin:
@7heo 

I'm a bit confused by your points here. This patch does not disable SSL verification unless `https_validate_certificates` is set to `False`, which is what you should reasonably expect from a setting called `https_validate_certificates`.

When `https_validate_certificates`, this doesn't disable SSL verification. That's what lines https://github.com/boto/boto/pull/2857/files#diff-d7ae0f691b0b4d12e00384a3132e19d8R167 and up do

Cheers,

7heo:
@krallin okay, my mistake, I reviewed too fast. I am going to edit the previous comment to reflect what I understood from your patch: https://github.com/boto/boto/pull/2857/files#diff-9d8becf47935de1157b07fe11f6b79daL750 is now broken in two lines, https://github.com/boto/boto/pull/2857/files#diff-9d8becf47935de1157b07fe11f6b79daR750 and https://github.com/boto/boto/pull/2857/files#diff-9d8becf47935de1157b07fe11f6b79daR751; which allows for a much better logic.

Sorry about the wrong review, and thanks for the heads up.

krallin:
@7heo 

That's fine — more eyes are always a good thing : ) I'm not sure this will ever be merged, but that's still useful to folks who might decide to pull in the patch.

Cheers,

jackar:
+ bump

suoinguon:
bump

PR #2858, opened 2015-01-05
Update LocaleRequirement to support subdivision

bob0the0mighty:
LocalRequirement upated to support subdivisions as an optional argument.  I'd be happy to create some tests too but it seems like the testing for the mturk module is a little haphazard so I'm not sure where to start.


awatts:
#3563 resolved the subdivision issue. This should probably be closed.

Issue #2870, opened 2015-01-09
Dynamo batch get infinite loop when it should return no results


adepue:
Due to https://github.com/boto/boto/pull/2809, if you query a table that should have no results, you end up in an infinite loop of requests trying to fetch more results.

PR #2871, opened 2015-01-09
Check for results left after computing self._keys_left

jlynn:
Addresses https://github.com/boto/boto/issues/2870

If the final call to fetch batch results returns no data,
BatchGetResultSet enters an infinite loop. Need to check that
self._keys_left is empty before returning.

### DIFF
--- a/boto/dynamodb2/results.py
+++ b/boto/dynamodb2/results.py
@@ -180,6 +180,9 @@ class BatchGetResultSet(ResultSet):
         kwargs['keys'] = self._keys_left[:self._max_batch_get]
         self._keys_left = self._keys_left[self._max_batch_get:]
 
+        if len(self._keys_left) <= 0:
+            self._results_left = False
+
         results = self.the_callable(*args, **kwargs)
 
         if not len(results.get('results', [])):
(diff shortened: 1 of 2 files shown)


boto/dynamodb2/results.py
```diff
@@ -180,6 +180,9 @@ def fetch_more(self):
         kwargs['keys'] = self._keys_left[:self._max_batch_get]
         self._keys_left = self._keys_left[self._max_batch_get:]
 
+        if len(self._keys_left) <= 0:
```
kereyroper: Is it necessary to invoke the_callable if len(_keys_left) <= 0? It seems like you should be able to return before then.

If _keys_left is [], then it seems like unprocessed_keys should always be [] later on, which would allow you to avoid the call and remove lines 199-200.

kereyroper:
It may be possible to simplify this change, but it seems to correct the bug, thanks. :+1:

adepue:
:+1: This corrects it for me as well

adepue:
@kyleknap @danielgtaylor @garthberry This is to correct a regression introduced in https://github.com/boto/boto/pull/2809 and released in 2.0.35.  Any chance to get it looked at soon?

rvarshney:
works for me too :+1:

asherf:
+1

amwhearsay:
+1

kyleknap:
Thanks for the PR. Looks good! Merging.

adepue:
Thanks @kyleknap! 
Any chance of a .1 release with the fix?

kyleknap:
@adepue I just did a patch release. The latest version is 2.35.1. Let me know if there are any other issues.

adepue:
Thanks!

garthberry:
@kereyroper Just caught up on this issue, I have submitted another pull request (#2880) that contains a simplified fix.

Issue #2872, opened 2015-01-09
cacerts.txt in inaccessible when boto is loaded using zipimport


petr-tichy:
I have packaged a script for standalone distribution as a zip file with boto included (http://pymotw.com/2/zipimport/), but it stopped working. The issue is due to inaccessible cacerts.txt file.

petr-tichy:
Closing this one as IMHO #3001 uses better solution

PR #2873, opened 2015-01-09
Fix cacerts.txt loading from within zipimport

petr-tichy:
Boto could not access the cacerts.txt file when loaded by zipimport from zip file. This solves #2872.
Please check if this is a sane solution as I'm python newbie.


petr-tichy:
Closing this one as IMHO #3001 uses better solution

PR #2880, opened 2015-01-13
Removed empty result set test to simplify fix for #2870

garthberry:
Removed empty result set test altogether, this short circuited the test for a completely empty result set and resulted in the infinite loop described in issue #2870

Removing the empty result set test allows the _keys_left test to correctly set _results_left in case of a completely empty result set.

The issue #2870 was already fixed by pull request #2871, this pull request just attempts to simplify the fix.


kereyroper:
looks good, thanks

PR #2884, opened 2015-01-14
Handle different platforms (Windows) with splitting BOTO_PATH

skelterjohn:
On Windows, paths are split with a ';' instead of a ':'. As a result, if you try to process BOTO_PATH on Windows, it will get something like ['C', '\path\to\file', 'D', '\more\files'] instead of ['C:\path\to\file', 'D:\more\files']. If windows is not provided with the drive name in the path, it will use the drive hosting your current working directory. So, if everything lives in C, including your terminal, it all coincidentally works. But if the file is in a different drive than your shell, it won't find it.

Using os.path.pathsep addresses this issue.


skelterjohn:
https://github.com/boto/boto/pull/2823/ already does this, please merge that one.

PR #2890, opened 2015-01-16
Switch to M2Crypto library

branto1:
python-rsa library exhibits several security flaws, this patch drops
the python-rsa dependency and implements the same functionality with
M2Crypto library.

Signed-off-by: Boris Ranto <branto@redhat.com>

### DIFF
--- a/boto/cloudfront/distribution.py
+++ b/boto/cloudfront/distribution.py
@@ -653,12 +653,13 @@ class Distribution(object):
         Requires the rsa library be installed.
         """
         try:
-            import rsa
+            from M2Crypto import EVP
         except ImportError:
-            raise NotImplementedError("Boto depends on the python rsa "
+            raise NotImplementedError("Boto depends on the M2Crypto "
                                       "library to generate signed URLs for "
                                       "CloudFront")
         # Make sure only one of private_key_file and private_key_string is set
+
         if private_key_file and private_key_string:
             raise ValueError("Only specify the private_key_file or the private_key_string not both")
         if not private_key_file and not private_key_string:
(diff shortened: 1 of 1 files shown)


branto1:
btw: I've tested this patch with the unittest tests/unit/cloudfront/test_signed_urls.py and the test passed.

danielgtaylor:
@branto1 thanks for the pull request. We are investigating the customer impact of an optional dependency switch as well as looking into the specific potential python-rsa issues to determine the best way to move forward.

branto1:
Thanks for the info @danielgtaylor .
btw: FWIW, a simple naive search at pkgs.org suggests that m2crypto is more likely to be available in a distro than python-rsa:
http://pkgs.org/search/m2crypto
http://pkgs.org/search/python-rsa

danielgtaylor:
@branto1 unfortunately M2Crypto doesn't seem to support Python 3, which Boto supports as of 2.32.0. We are still investigating the best way to move forward with this, taking into account Python version support and customer impact.

mcepl:
I would strongly suggest AGAINST using M2Crypto. I am one of the few co-maintainers of it (and my branch https://github.com/mcepl/m2crypto/tree/python3 is AFAIK the only effort to port M2Crypto to Python 3) and I can see that it is not in the good shape. I would strongly suggest switching to some other bindings (of course, using home-brewn crypto is a suicide, you should use bindings to some popular library, that's true), namely https://github.com/pyca/cryptography/ seems to be quite popular these days.

alex:
*waves* I work on pyca/cryptography, if there's anything we can do to smooth out adoption, please let us know.

branto1:
Actually, I think that the simplest solution here would be to use pyOpenSSL directly, it supports python2 as well as python3, it is fast, fairly widely used and it supports signing for quite some time (rhel 7 package does have that capability, rhel 6 does not). In fact, I've created pull request for that effort:

https://github.com/boto/boto/pull/2930

mcepl:
> Actually, I think that the simplest solution here would be to use pyOpenSSL directly

Just to note that the later versions of pyOpenSSL are actually front-ends for python-cryptography.

mcepl:
> Actually, I think that the simplest solution here would be to use pyOpenSSL directly

Just to note that the later versions of pyOpenSSL are actually front-ends for python-cryptography.

grahamlyons:
Has there been any progress on this?

PR #2891, opened 2015-01-16
Kinesis update

kyleknap:
This updates Kinesis to the latest API. This includes being able to tag streams and use ``put_records``. I also pulled in the commits from another PR from @danielgtaylor that is still open: https://github.com/boto/boto/pull/2555. That was needed so that binary data for both ``put_record`` and ``put_records`` were treated properly.

cc @danielgtaylor @jamesls

### DIFF
--- a/boto/kinesis/layer1.py
+++ b/boto/kinesis/layer1.py
@@ -28,6 +28,7 @@ from boto.regioninfo import RegionInfo
 from boto.exception import JSONResponseError
 from boto.kinesis import exceptions
 from boto.compat import json
+from boto.compat import six
 
 
 class KinesisConnection(AWSQueryConnection):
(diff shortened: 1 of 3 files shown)


boto/kinesis/layer1.py
```diff
@@ -591,66 +653,171 @@ def put_record(self, stream_name, data, partition_key,
...
+        :param b64_encode: Whether to Base64 encode `data`. Can be set to
+            ``False`` if `data` is already encoded to prevent double encoding.
+
+        """
+        params = {'Records': records, 'StreamName': stream_name, }
+        if b64_encode:
+            for i in range(len(params['Records'])):
+                data = params['Records'][i]['Data']
+                if not isinstance(data, six.binary_type):
+                    params['Records'][i]['Data'] = data.encode('utf-8')
```
danielgtaylor: Should this be `data = data.encode('utf-8')`? Otherwise we are never using the encoded value.

kyleknap: Good catch. It is failing the integration test that I wrote for py3. I will fix it.

kyleknap:
@danielgtaylor I made the fix in encoding for ``put_records()``. Ready to get looked at again.

danielgtaylor:
LGTM :ship:-it!

Issue #2869, opened 2015-01-08
SSL over proxy is broken in Python 3.4


Raniz-at-Sony:
I don't know if these were changed between 3.3 and 3.4, but it's not working for me though:

* *Socket.sendall* requires bytes as input and not strings.
* *HTTPResponse* no longer supports the "strict" parameter.

As far as I can tell, everything is in connection.py in the *proxy_ssl* method.

tiznull:
You can pull a fox I posted for this same issue.

https://github.com/tiznull/boto/compare/boto:develop...patch-1?quick_pull=1

AkshivBaluja:
Thanks for the fix ! , it has allowed me to use boto for python 3.4 over secure proxy !

PR #2893, opened 2015-01-16
Update connection.py

tiznull:
We had the same issue as report in ticket [boto]#2869.

These are the problems that this update fixes:
* Socket.sendall requires bytes as input and not strings.
* HTTPResponse no longer supports the "strict" parameter.


tiznull:
I am working on the build errors, I will submit a new PULL request once those are corrected.

danielgtaylor:
@tiznull can you give more details about when you encounter errors related to this pull request?

PR #2894, opened 2015-01-16
KMS Support

kyleknap:
This adds support for KMS. Note that any input argument that is modeled as a blob must be in bytes (it gets validated before making the request).

cc @danielgtaylor @jamesls

### DIFF
--- a/boto/kms/layer1.py
+++ b/boto/kms/layer1.py
@@ -274,12 +274,11 @@ class KMSConnection(AWSQueryConnection):
             can be used to provide long term permissions to perform decryption.
 
         """
-        if ciphertext_blob is not None:
-            if not isinstance(ciphertext_blob, six.binary_type):
-                raise TypeError(
-                    "Argument ``ciphertext_blob`` is "
-                    "modeled as a blob. Value must be bytes.")
-            ciphertext_blob = base64.b64encode(ciphertext_blob)
+        if not isinstance(ciphertext_blob, six.binary_type):
+            raise TypeError(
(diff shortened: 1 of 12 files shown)


boto/kms/layer1.py
```diff
@@ -0,0 +1,824 @@
...
+        :param encryption_context: Name:value pair that specifies the
+            encryption context to be used for authenticated encryption. For
+            more information, see `Authenticated Encryption`_.
+
+        :type grant_tokens: list
+        :param grant_tokens: A list of grant tokens that represent grants which
+            can be used to provide long term permissions to perform encryption.
+
+        """
+        if plaintext is not None:
```
jamesls: Is there a reason for the `is not None` check?  `plaintext` is a required param.

boto/kms/layer1.py
```diff
@@ -0,0 +1,824 @@
...
+            if not isinstance(plaintext, six.binary_type):
+                raise TypeError(
+                    "Argument ``plaintext`` is "
+                    "modeled as a blob. Value must be bytes.")
```
jamesls: I suspect users may not understand the `modeled as a blob` bit.  What if we just say something like "value must be of type %s" % six.binary_type?

jamesls:
Looks good, just a question about the is not None check for required params.

kyleknap:
@jamesls I updated the code with your suggestions. Let me know what you think.

jamesls:
:shipit: Looks good.

Issue #2864, opened 2015-01-07
Adding support for AWS Lambda?


Captainkrtek:
Hi,
I recently gained access to the AWS Lambda preview and was interested in starting some groundwork on adding support for it to boto. I was wondering if this would be accepted, is already in the works, or if we are holding off until it is no longer a preview?
Thoughts?

Captainkrtek:
@garnaat @kyleknap ?

PR #2896, opened 2015-01-19
AWS Lambda

kyleknap:
This adds layer1 support for AWS Lambda. Note that I use awslambda as opposed to lambda because lambda is a keyword in python.

Fixes https://github.com/boto/boto/issues/2864

cc @danielgtaylor @jamesls


kyleknap:
@jamesls I added better error checking for unseekable streams.

Ready to get looked at again.

boto/awslambda/__init__.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) 2014 Amazon.com, Inc. or its affiliates.  All Rights Reserved
```
danielgtaylor: 2015 now :smile:

danielgtaylor:
Aside from the 2015 year update, LGTM :ship:-it!

PR #2898, opened 2015-01-20
add filter support for boto.ec2.autoscale.get_all_tags()

WrathOfChris:
Implements the filters parameter for boto.ec2.autoscale.get_all_tags()

Valid filter keys are:
* `auto-scaling-group`
* `key`
* `value`
* `propagate-at-launch`

(ref: http://docs.aws.amazon.com/AutoScaling/latest/APIReference/API_Filter.html)

### DIFF
--- a/tests/unit/ec2/autoscale/test_group.py
+++ b/tests/unit/ec2/autoscale/test_group.py
@@ -603,6 +603,54 @@ class TestAutoScalingTag(AWSMockServiceTestCase):
         else:
             self.assertEqual(getattr(tag, attr), value)
 
+class TestAutoScalingTagFilter(AWSMockServiceTestCase):
+    connection_class = AutoScaleConnection
+
+    def default_body(self):
+        return b"""
+        <DescribeTagsResponse>
+            <ResponseMetadata>
+                <RequestId>requestId</RequestId>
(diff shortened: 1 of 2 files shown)


tests/unit/ec2/autoscale/test_group.py
```diff
@@ -603,6 +603,54 @@ def check_tag_attributes_set(self, name, value, attr):
...
+            <ResponseMetadata>
+                <RequestId>requestId</RequestId>
+            </ResponseMetadata>
+        </DescribeTagsResponse>
+        """
+
+    def test_get_all_tags(self):
+        self.set_http_response(status_code=200)
+
+        tags = [
```
danielgtaylor: What's the purpose of making this list of tag objects? They don't seem to get used anywhere.

WrathOfChris: Ah, you're right!  Call it a consequence of too much copy-paste as I learned how to write a boto test.  Cheers.

WrathOfChris:
The order of the filters wasn't consistent between travis runs - is there a better way to test for the 2 filters being present, without requiring the ordering to match as `assert_request_parameters` seems to require?

PR #2906, opened 2015-01-22
Add sts for eu-central-1

clifflu:
Endpoint for STS in eu-central-1 was missing, returning None on legit calls.


danielgtaylor:
Thanks :+1:

PR #2908, opened 2015-01-22
Add support for Amazon EC2 Container Service

kyleknap:
This PR adds support for EC2 Container Service. Note that I was not able to call it ``ecs`` because of a previous ``ecs``, which is the AWS Ecommerce Service. I decided to name it ``ec2containerservice`` with ``EC2ContainerServiceConnection`` objects. Even though it is a longer name (it is 1 or 2 characters longer than the longest module/connection name), it is more explicit, which hopefully will lessen confusion between the two services in boto.

I also looked into adding ECS to the ``ec2`` directory and calling it ``ecs`` with ``ECSConnection`` objects however there was still a naming collision with the AWS Ecommerce Service's ``ECSConnection`` and there would be some possible naming collisions in the docs and possible confusion in the ``endpoints.json`` with the Ecommerce service even though it does not have any entry in the ``endpoints.json``.

cc @danielgtaylor @jamesls

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -35,6 +35,7 @@ At the moment, boto supports:
   * AutoScaling (Python 3)
   * Amazon Kinesis (Python 3)
   * AWS Lambda (Python 3)
+  * Amazon EC2 Container Service (Python 3)
 
 * Content Delivery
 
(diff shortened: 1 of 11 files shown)


danielgtaylor:
Just update the dates and then LGTM :ship:-it!

kyleknap:
I updated the copyright year to 2015. Merging.

Issue #2865, opened 2015-01-07
Message can not be sent into SQS queue in 2.34 version if the message body is not utf-8 encodable


pengfeicui:
My python version is 2.7.5, boto version is 2.34.0, and test script is:

    import zlib

    import boto.sqs

    conn = boto.sqs.connect_to_region('us-west-2')
    q = conn.get_queue('my-test-queue')
    m = boto.sqs.message.Message()
    m.set_body(zlib.compress('This is my test message.'))
    q.write(m)
I write some strings (not utf-8 encodable) in SQS message body, but in 2.34.0, the encode function of class Message() encodes such strings into utf-8 string. So bellow exception is raised, and the traceback message is:

Traceback (most recent call last):
  File "test.py", line 10, in <module>
    q.write(m)
  File "/usr/local/lib/python2.7/site-packages/boto-2.34.0-py2.7.egg/boto/sqs/queue.py", line 225, in     write
    message.get_body_encoded(), delay_seconds=delay_seconds,
  File "/usr/local/lib/python2.7/site-packages/boto-2.34.0-py2.7.egg/boto/sqs/message.py", line     141, in get_body_encoded
    return self.encode(self.get_body())
  File "/usr/local/lib/python2.7/site-packages/boto-2.34.0-py2.7.egg/boto/sqs/message.py", line    166, in encode
    return base64.b64encode(value.encode('utf-8')).decode('utf-8')
UnicodeDecodeError: 'ascii' codec can't decode byte 0x9c in position 1: ordinal not in range(128)

After checking the source code of boto/sqs/message.py, i found that class Message() has some change in class function of encode(self, value) and decode(self, value) between 2.30 and 2.34.

In 2.34:

    class Message(RawMessage):
        def encode(self, value):
            return base64.b64encode(value.encode('utf-8')).decode('utf-8')

        def decode(self, value):
            try:
                value = base64.b64decode(value.encode('utf-8')).decode('utf-8')
            except:
                boto.log.warning('Unable to decode message')
                return value
            return value

In 2.30:

    class Message(RawMessage):
        def encode(self, value):
            return base64.b64encode(value)

        def decode(self, value):
             try:
                value = base64.b64decode(value)
            except:
                boto.log.warning('Unable to decode message')
                return value
            return value

I think SQS should not care the character type in the message body, like the 2.30 does. But it seems only utf-8 encodeable strings can use the SQS in 2.34.

[...]

PR #2913, opened 2015-01-23
Allow for binary to be passed to sqs message

kyleknap:
Fixes https://github.com/boto/boto/issues/2865

Originally was not taking into account binary data when base64 encoding the message body.

cc @danielgtaylor @jamesls


jamesls:
:shipit:

Issue #2917, opened 2015-01-27
RDS snapshot can't be restored in different VPC than the original specifying a db_subnet_group_name when RDS option group is associated to the original VPC


ronaldosaheki:
RDS snapshot can't be restored to the last snapshot using the `restore_dbinstance_from_point_in_time` method in `RDSConnection` class, to a different VPC than the original (specifying a db_subnet_group_name in a different VPC) when RDS option group is associated to the original VPC.

It raises the following exception:
```
File "/usr/local/lib/python2.7/dist-packages/boto-2.34.0-py2.7.egg/boto/rds/__init__.py", line 1338, in restore_dbinstance_from_point_in_time
    params, DBInstance)
  File "/usr/local/lib/python2.7/dist-packages/boto-2.34.0-py2.7.egg/boto/connection.py", line 1204, in get_object
    raise self.ResponseError(response.status, response.reason, body)
boto.exception.BotoServerError: BotoServerError: 400 Bad Request
<ErrorResponse xmlns="http://rds.amazonaws.com/doc/2013-05-15/">
  <Error>
    <Type>Sender</Type>
    <Code>InvalidParameterCombination</Code>
    <Message>The option group stage-options-oracle-se1 is associated with a different VPC than the request.</Message>
  </Error>
  <RequestId>730760fb-a62d-11e4-aa94-95370f30c8ec</RequestId>
</ErrorResponse>
```

PR #2919, opened 2015-01-27
Fixes boto/boto#2917 and updated dbinstance_class parameter comment and instance_class parameter name to dbinstance_class

ronaldosaheki:
Fixes boto/boto#2917 and updated `dbinstance_class` parameter comment and changed the `instance_class` parameter name in `restore_dbinstance_from_dbsnapshot`  to `dbinstance_class` to let the parameter name the same in both functions `restore_dbinstance_from_point_in_time` and `restore_dbinstance_from_dbsnapshot`


danielgtaylor:
@ronaldosaheki thanks for the pull request. Typically new work will happen on the `rds2` module, which is code-generated rather than written by hand. If you still wish to make these changes to provide support for people using the `rds` module then you will need to add some unit tests that cover the new functionality. Aside from that, this change looks good to me :+1:

ronaldosaheki:
Unit tests adjusted, it's just that most of the documentation use rds instead of rds2. At least for basic it should work.

ronaldosaheki:
Added new unit test to cover added funcionality: restore a database snapshot from point in time to a new RDS instance into another VPC, specifying a subnetgroup and option-group from target VPC.

ronaldosaheki:
@danielgtaylor Test to cover functionality added

PR #2920, opened 2015-01-28
multipart_upload_lister: support for prefix

jraby:
`prefix` is simply passed to `get_all_multipart_uploads()`. 
This allows to get prefix filtered multipart uploads much faster than filtering the full results on the client side.


hannes-ucsc:
+1

hannes-ucsc:
This seems to be a trivial change. Anything stopping it from being merged?

PR #2924, opened 2015-01-30
DynamoDBConnection handles host=None

bobbyi:
The default value of the host argument to DynamoDBConnection's constructor is documented as None because it subclasses AWSQueryConnection. However, explicitly passing host=None did not behave the same as not passing a value because the constructor checked for 'host' as key in kwargs.

### DIFF
--- a/boto/dynamodb2/layer1.py
+++ b/boto/dynamodb2/layer1.py
@@ -169,7 +169,7 @@ class DynamoDBConnection(AWSQueryConnection):
                     break
 
         # Only set host if it isn't manually overwritten
-        if 'host' not in kwargs:
+        if kwargs.get('host') is None:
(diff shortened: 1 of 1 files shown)


boto/dynamodb2/layer1.py
```diff
@@ -169,7 +169,7 @@ def __init__(self, **kwargs):
                     break
 
         # Only set host if it isn't manually overwritten
-        if 'host' not in kwargs:
+        if kwargs.get('host') is None:
             kwargs['host'] = region.endpoint
```
xethorn: `kwargs['host'] = kwargs.get('host') or region.endpoint` # so at least it's consistently defined – and if someone passes 0 or False, it will behave the same.

PR #2925, opened 2015-01-30
Add Amazon DynamoDB online indexing support on High level API

pauloubuntu:
Hi,

Yesterday (Jan-29), Amazon has added support to add or delete global indexes at any time, which is by far the feature I was most looking forward using for over a year. 

I've changed dynamodb2.table.update method taking into account the backward campatibility as well. Documentation amendments and a unit test case has been provided too accordingly to the contributing guidelines.

Looking forward to see this in the new version of boto.

### DIFF
--- a/boto/dynamodb2/table.py
+++ b/boto/dynamodb2/table.py
@@ -371,10 +371,9 @@ class Table(object):
             raw_indexes = result['Table'].get('LocalSecondaryIndexes', [])
             self.indexes = self._introspect_indexes(raw_indexes)
 
-        if not self.global_indexes:
-            # Build the global index information as well.
-            raw_global_indexes = result['Table'].get('GlobalSecondaryIndexes', [])
-            self.global_indexes = self._introspect_global_indexes(raw_global_indexes)
+        # Build the global index information as well.
+        raw_global_indexes = result['Table'].get('GlobalSecondaryIndexes', [])
+        self.global_indexes = self._introspect_global_indexes(raw_global_indexes)
 
         # This is leaky.
         return result
@@ -465,6 +464,9 @@ class Table(object):
         Requires a ``global_indexes`` parameter, which should be a
         ``GlobalBaseIndexField`` subclass representing the desired index.
 
+        To update ``global_indexes`` information on the ``Table``, you'll need
+        to call ``Table.describe``.
+
         Returns ``True`` on success.
 
         Example::
@@ -517,6 +519,9 @@ class Table(object):
         Requires a ``global_index_name`` parameter, which should be a simple
         string of the name of the global secondary index.
 
+        To update ``global_indexes`` information on the ``Table``, you'll need
+        to call ``Table.describe``.
+
         Returns ``True`` on success.
 
         Example::
@@ -557,6 +562,9 @@ class Table(object):
         dictionary. If provided, it should specify the index name, which is also
         a dict containing a ``read`` & ``write`` key, both of which
         should have an integer value associated with them.
+
+        To update ``global_indexes`` information on the ``Table``, you'll need
+        to call ``Table.describe``.
 
         Returns ``True`` on success.
 
--- a/boto/dynamodb2/table.py
+++ b/boto/dynamodb2/table.py
@@ -443,12 +443,20 @@ class Table(object):
                     },
                 })
 
-        self.connection.update_table(
-            self.table_name,
-            provisioned_throughput=data,
-            global_secondary_index_updates=gsi_data,
-        )
-        return True
+        if throughput or global_indexes:
+            self.connection.update_table(
+                self.table_name,
+                provisioned_throughput=data,
+                global_secondary_index_updates=gsi_data,
+            )
+
+            return True
+        else:
+            msg = 'You need to provide either the throughput or the ' \
+                  'global_indexes to update method'
+            boto.log.error(msg)
+
+            return False
 
     def create_global_secondary_index(self, global_index):
         """
@@ -477,9 +485,6 @@ class Table(object):
 
         """
 
-        gsi_data = None
-        gsi_data_attr_def = None
-
         if global_index:
             gsi_data = []
             gsi_data_attr_def = []
@@ -491,12 +496,19 @@ class Table(object):
             for attr_def in global_index.parts:
                 gsi_data_attr_def.append(attr_def.definition())
 
-        self.connection.update_table(
-            self.table_name,
-            global_secondary_index_updates=gsi_data,
-            attribute_definitions=gsi_data_attr_def
-        )
-        return True
+            self.connection.update_table(
+                self.table_name,
+                global_secondary_index_updates=gsi_data,
+                attribute_definitions=gsi_data_attr_def
+            )
+
+            return True
+        else:
+            msg = 'You need to provide the global_index to ' \
+                  'create_global_secondary_index method'
+            boto.log.error(msg)
+
+            return False
 
     def delete_global_secondary_index(self, global_index_name):
         """
@@ -515,8 +527,6 @@ class Table(object):
 
         """
 
-        gsi_data = None
-
         if global_index_name:
             gsi_data = [
                 {
@@ -526,11 +536,18 @@ class Table(object):
                 }
             ]
 
-        self.connection.update_table(
-            self.table_name,
-            global_secondary_index_updates=gsi_data,
-        )
-        return True
+            self.connection.update_table(
+                self.table_name,
+                global_secondary_index_updates=gsi_data,
+            )
+
+            return True
+        else:
+            msg = 'You need to provide the global index name to ' \
+                  'delete_global_secondary_index method'
+            boto.log.error(msg)
+
+            return False
 
     def update_global_secondary_index(self, global_index):
         """
@@ -555,7 +572,6 @@ class Table(object):
             True
 
         """
-        gsi_data = None
 
         if global_index:
             gsi_data = []
@@ -571,11 +587,17 @@ class Table(object):
                     },
                 })
 
-        self.connection.update_table(
-            self.table_name,
-            global_secondary_index_updates=gsi_data,
-        )
-        return True
+            self.connection.update_table(
+                self.table_name,
+                global_secondary_index_updates=gsi_data,
+            )
+            return True
+        else:
+            msg = 'You need to provide the global index to ' \
+                  'update_global_secondary_index method'
+            boto.log.error(msg)
+
+            return False
 
     def delete(self):
         """
--- a/boto/dynamodb2/table.py
+++ b/boto/dynamodb2/table.py
@@ -383,17 +383,15 @@ class Table(object):
         """
         Updates table attributes and global indexes in DynamoDB.
 
-        Now, you can modify table's throughput , create/update/delete global
-        indexes at any time.
-
         Optionally accepts a ``throughput`` parameter, which should be a
         dictionary. If provided, it should specify a ``read`` & ``write`` key,
         both of which should have an integer value associated with them.
 
         Optionally accepts a ``global_indexes`` parameter, which should be a
-        dictionary. If provided, it should specify a different key structure
-        depending on the operation you want to perform ( ``create``, ``update``,
-        ``delete``).
+        dictionary. If provided, it should specify the index name, which is also
+        a dict containing a ``read`` & ``write`` key, both of which
+        should have an integer value associated with them. If you are writing
+        new code, please use ``Table.update_global_secondary_index``.
 
         Returns ``True`` on success.
 
@@ -412,34 +410,12 @@ class Table(object):
             ...     'write': 10,
             ... },
             ... global_secondary_indexes={
-            ...     'update': {
-            ...         'TheIndexNameHere': {
-            ...             'read': 15,
-            ...             'write': 5,
-            ...         }
-            ...      }
-            ... })
-
-            # To create a global index(es)
-            >>> users.update(global_secondary_indexes={
-            ...     'create': {
-            ...         GlobalAllIndex('TheIndexNameHere', parts=[
-            ...             HashKey('requiredHashkey', data_type=STRING),
-            ...             RangeKey('optionalRangeKey', data_type=STRING)
-            ...         ],
-            ...         throughput={
-            ...             'read': 2,
-            ...             'write': 1,
-            ...         }),
-            ...      }
-            ... })
-
-            # To delete a global index(es)
-            >>> users.update(global_secondary_indexes={
-            ...     'delete': 'TheIndexNameHere'
+            ...     'TheIndexNameHere': {
+            ...         'read': 15,
+            ...         'write': 5,
+            ...     }
             ... })
             True
-
         """
 
         data = None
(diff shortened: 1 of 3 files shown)


boto/dynamodb2/table.py
```diff
@@ -379,17 +379,22 @@ def describe(self):
...
 
-    def update(self, throughput, global_indexes=None):
+    def update(self, throughput=None, global_indexes=None):
         """
-        Updates table attributes in DynamoDB.
+        Updates table attributes and global indexes in DynamoDB.
 
-        Currently, the only thing you can modify about a table after it has
-        been created is the throughput.
+        Now, you can modify table's throughput , create/update/delete global
```
xethorn: Might not be useful to mention `Now`.

boto/dynamodb2/table.py
```diff
@@ -379,17 +379,22 @@ def describe(self):
         # This is leaky.
         return result
 
-    def update(self, throughput, global_indexes=None):
+    def update(self, throughput=None, global_indexes=None):
```
xethorn: The responsibility of this method is to update table attributes. 

Here, the addition allows the creation, update and deletion of GSI. Then it seems like those changes belong in different (potentially smaller) methods. It will keep the complexity of this method to the minimum.

What do you think? @danielgtaylor

boto/dynamodb2/table.py
```diff
@@ -407,39 +412,100 @@ def update(self, throughput, global_indexes=None):
...
+            ...         'TheIndexNameHere': {
+            ...             'read': 15,
+            ...             'write': 5,
+            ...         }
+            ...      }
+            ... })
+
+            # To create a global index(es)
+            >>> users.update(global_secondary_indexes={
+            ...     'create': {
```
xethorn: To follow the previous comment, this will just end up being:

```create_global_secondary_index(name_of_index, parts, throughput)```

File: boto/dynamodb2/table.py [hunk 4]
```diff
@@ -407,39 +412,100 @@ def update(self, throughput, global_indexes=None):
...
+                'WriteCapacityUnits': int(self.throughput['write']),
+            }
+
         gsi_data = None
+        gsi_data_attr_def = None
 
         if global_indexes:
             gsi_data = []
 
-            for gsi_name, gsi_throughput in global_indexes.items():
```
xethorn: this code will just end up using the new `create_global_secondary_index`

boto/dynamodb2/table.py
```diff
@@ -407,39 +412,100 @@ def update(self, throughput, global_indexes=None):
...
-                gsi_data.append({
-                    "Update": {
-                        "IndexName": gsi_name,
-                        "ProvisionedThroughput": {
-                            "ReadCapacityUnits": int(gsi_throughput['read']),
-                            "WriteCapacityUnits": int(gsi_throughput['write']),
+            for gsi_operation_key, gsi_operation_value in global_indexes.items():
+                if gsi_operation_key == 'delete':
+
+                    gsi_data.append({
```
xethorn: `delete_global_secondary_index` maybe?

boto/dynamodb2/table.py
```diff
@@ -407,39 +412,100 @@ def update(self, throughput, global_indexes=None):
...
+                        "Delete": {
+                            "IndexName": gsi_operation_value
+                        },
+                    })
+
+                elif gsi_operation_key == 'update':
+
+                    for gsi_update_key, gsi_throughput in gsi_operation_value.items():
```
xethorn: Same here

boto/dynamodb2/table.py
```diff
@@ -407,39 +412,100 @@ def update(self, throughput, global_indexes=None):
...
+                    gsi_data.append({
+                        "Create": gsi_operation_value.schema()
+                    })
+
+                    gsi_data_attr_def = []
+
+                    for attr_def in gsi_operation_value.parts:
+                        gsi_data_attr_def.append(attr_def.definition())
+                else:
+                    # Provide backward compatibility for the old dict structure
```
xethorn: You wouldn't have to provide backward compatibility anymore. But: it would be good to keep the comment you wrote, for reference.

pauloubuntu:
@xethorn Wow, so many comments at once. It’s really cool to see I'm not the only guy interested on this DynamoDB feature :)

xethorn:
@pauloubuntu : this is really looking great! I've added minor comments (it's mostly about keep the methods as small as possible to keep its complexity to the bare minimum.) – those comments are definitely not blocking if @danielgtaylor is fine with the changes!

xethorn:
@pauloubuntu : This change is definitely super exciting :+1:

pauloubuntu:
@xethorn 

I've actually changed update method because it calls UpdateTableAPI which is the same endpoint the aws-cli uses to create/update/delete global secondary indexes. When I was first implementing this feature I thought I should keep it in that way. However, I agree that there are a couple of points that can be improved and I’m really willing to do it. 

Thanks for all the code line comments you've made.

I hope I can get 'em done as soon as I get home 

Cheers

pauloubuntu:
I really liked the idea of creating delete_global_secondary_index, update_global_secondary_index, create_global_secondary_index methods instead of keeping 'em under the update method. Once again thanks for pointing this out.

xethorn:
:+1: !

danielgtaylor:
@pauloubuntu, @xethorn I like the idea of separate methods, and this is both simpler to call and seems consistent with other pieces of Boto :+1: Aside from that change this looks very good. The only other comment I have is maybe to add some unit tests, since the integration test suite tends to get ignored by a lot of contributors and failing during unit tests would be preferable as this is a module with a lot of use and a lot of contributors.

tests/integration/dynamodb2/test_highlevel.py
```diff
@@ -716,3 +716,68 @@ def test_query_after_describe_with_gsi(self):
 
         for rs_item in rs:
             self.assertEqual(rs_item['username'], ['johndoe'])
+
+    def test_update_table_online_indexing_support(self):
```
xethorn: @danielgtaylor : are we running those tests against a local dynamodb?

xethorn: @pauloubuntu : nothing to do here, i'm just curious.

danielgtaylor: @xethorn, these tests are run against the live DynamoDB service using your configured credentials. It's possible that you could run against a DynamoDB Local service, but it is not used by default.

xethorn: Got it. Thanks!

pauloubuntu:
Hi, @xethorn , @danielgtaylor 

I've splitted the logic in smaller methods as we've discussed before. Here goes a few observations:

- I've reverted the update method to the previous implementation and added a note for future developers that if they're writing a new code then they should be using update_global_secondary_index method instead
- On create_global_secondary_index I've decided to use GlobalBaseIndexField subclass as a parameter since it has already all the paramters we need in a well-know form instead of breaking its attributes in three different parameters such as : 
```Python
def create_global_secondary_index(self, index_name, parts, throughput)
```

Let me know if you guys want to change anything while I'm coding the unit tests.

Cheers

boto/dynamodb2/table.py
```diff
@@ -439,7 +446,134 @@ def update(self, throughput, global_indexes=None):
...
+            ...         })
+            ... )
+            True
+
+        """
+
+        gsi_data = None
+        gsi_data_attr_def = None
+
+        if global_index:
```
xethorn: The global index should not be null – otherwise you shouldn't update the table. Not sure what @danielgtaylor would prefer for this, but it seems like `if not global_index`, just return False and log the fact that it's empty?

boto/dynamodb2/table.py
```diff
@@ -439,7 +446,134 @@ def update(self, throughput, global_indexes=None):
...
+
+            # To delete a global index
+            >>> users.delete_global_secondary_index('TheIndexNameHere')
+            True
+
+        """
+
+        gsi_data = None
+
+        if global_index_name:
```
xethorn: Same here: `global_index_name` is a requirement for this method to fully work. So if empty: `return False` and add a line in the logs ("You need to provide the index name to delete_global_secondary_index", or something along those lines.)

boto/dynamodb2/table.py
```diff
@@ -439,7 +446,134 @@ def update(self, throughput, global_indexes=None):
...
+            ...         'read': 15,
+            ...         'write': 5,
+            ...     }
+            ... })
+            True
+
+        """
+        gsi_data = None
+
+        if global_index:
```
xethorn: Same applies here.

xethorn:
This looks even better than the previous one! Almost there. Let me know if you have any questions :)

pauloubuntu:
@xethorn what about now ? :)

pauloubuntu:
@xethorn  @danielgtaylor 
Actually. I was little confused about using boto.log.error in this case. Should I use warn instead ?

What do you think ?

xethorn:
Haven't used boto.log. @danielgtaylor will know better.

Code looks great! :+1:

tests/integration/dynamodb2/test_highlevel.py
```diff
@@ -716,3 +716,68 @@ def test_query_after_describe_with_gsi(self):
...
+        users.update_global_secondary_index(global_index={
+            'EmailGSIIndex': {
+                'read': 2,
+                'write': 1,
+            }
+        })
+
+        # Wait for it.
+        time.sleep(60)
+
```
kyleknap: For tests, we typically like it if you explicitly assert something as opposed to just testing that you can successfully call the method. I would suggest using the table's layer2 to make a call to ``describe_table()`` and then asserting that the response has the global secondary indexes that you are expecting.

pauloubuntu: Got it. 

Thanks for pointing this out.

boto/dynamodb2/table.py
```diff
@@ -436,12 +443,161 @@ def update(self, throughput, global_indexes=None):
...
+                  'global_indexes to update method'
+            boto.log.error(msg)
+
+            return False
+
+    def create_global_secondary_index(self, global_index):
+        """
+        Creates a global index in DynamoDB after the table has been created.
+
+        Requires a ``global_indexes`` parameter, which should be a
```
kyleknap: Should be ``global_index`` in the doc string.

boto/dynamodb2/table.py
```diff
@@ -436,12 +443,161 @@ def update(self, throughput, global_indexes=None):
...
+
+            return True
+        else:
+            msg = 'You need to provide the global index name to ' \
+                  'delete_global_secondary_index method'
+            boto.log.error(msg)
+
+            return False
+
+    def update_global_secondary_index(self, global_index):
```
kyleknap: It seems a little strange to have the ``update_global_secondary_index()`` method as well as an ``update()`` method since the ``update()`` method does the ``update_global_secondary_index()`` method's functionality plus throughput updating. But I suppose it would be more strange (consistency-wise) if there are ``create_global_secondary_index()`` and  ``delete_global_secondary_index()`` methods, but no corresponding ``update_global_secondary_index()`` method.

As to the implementation of the method could you just make it:
```
def update_global_secondary_index(self, global_index):
    return self.update(global_indexes=global_index)
```
That would help simplify the code and reuse some.


On a side note, it looks like this method can update multiple indexes at once. Should you update the name of the parameter ``global_index`` to ``global_indexes`` especially since you use ``global_indexes`` in the doc string?

kyleknap:
Thanks for the PR! It looks good to me. I just had a few comments.

The only other general comment that I had was similar to the @danielgtaylor comment, which was to add some unit tests because those are ran more than the integration tests. If you need help writing some, I would suggest following the style here:
https://github.com/boto/boto/blob/develop/tests/unit/dynamodb2/test_table.py#L1519

where you patch the layer1's ``update_table`` method and make sure that the correct parameters were passed to the patched method.

boto/dynamodb2/table.py
xethorn: Maybe we should deprecate `update`?

It might be easier to be able to update the structure of the table using smaller methods, then save those changes? So you could chain: adding new indexes, and removing some others – all in one call.

If in the future dynamodb adds more features (e.g. be able to change the name of the table, be able to create/update/remove local secondary indexes), then this method will end up quite large and more difficult to test.

That's definitely outside of the scope of this pull request tho – but if that approach is a good idea, then maybe we should not make `update_global_secondary_index` point to `update`?

pauloubuntu:
@kyleknap I've changed that parameter to global_indexes and I've also added a few assertions on test_highlevel.py. Just a few observations and one question:

- I've used the highlevel describe() method to retrieve and assert if everything is working as expected instead of the layer1 since I couldn't find usages of any method that wasn't from the highlevel api. So I'm assuming this integration should only use highlevel api methods. ( Correct if I'm wrong please :) )

- I haven't touched the update_global_secondary_index internal logic since we are arguing which should be the best approach for this case. I agree with @xethorn that we should keep it as a separate method and the use of update method to update global_indexes throughput. This is going to make things a bit easier as DynamoDB continues to add new features. Let me know your opinion about this.

- I'm still working on some unit tests which I hope to finish in a few hours.

Cheers

kyleknap:
@pauloubuntu Awesome! Thanks for seeing this through. Changes that you have made so far look good.

Also @pauloubuntu @xethorn that is fine if we do not share code between the ``update()`` and ``update_global_secondary_index()`` method. I was just a suggesting a way to save you some code/logic.

I agree that we should not be bloating the ``update()`` method with more features. This is will have to be for a PR later down the road, but we cannot begin deprecating the ``update()`` method till a ``Table`` level method is added that can update throughput because I believe the ``update()`` method is the only function that can, not including the layer1 methods.

pauloubuntu:
@kyleknap , @xethorn , @danielgtaylor I've just pushed a few unit tests to test_table.py following the style suggested. 

Before you check it I need to ask you a question. As @kyleknap said, unit tests are supposed to explicitly assert something, but to assert that I have created/updated/deleted any online global indexes I should update the self.global_indexes variable which wasn't been updated by neither the update method nor the newest methods we've created.  I have also noticed that the describe method only updates the self.global_indexes if it's None so I should remove that condition as well.

( I'll write just a method to give an example)
The create method would be something like it:

```Python
def create_global_secondary_index(self, global_index):
        if global_index:
            # Add a GSI to global_indexes
            self.global_indexes.append(global_index)

            gsi_data = []
            gsi_data_attr_def = []

            gsi_data.append({
                "Create": global_index.schema()
            })

            for attr_def in global_index.parts:
                gsi_data_attr_def.append(attr_def.definition())

            self.connection.update_table(
                self.table_name,
                global_secondary_index_updates=gsi_data,
                attribute_definitions=gsi_data_attr_def
            )

            return True
        else:
            msg = 'You need to provide the global_index to ' \
                  'create_global_secondary_index method'
            boto.log.error(msg)

            return False
```

What do you think about it ?

kyleknap:
The unit test look great. You are explicitly asserting something in those tests. You are asserting what the connection mock was called with, which is all the testing needed for this abstraction.

For actually updating the ``Table`` instances when calling those new methods you added, I would say do not update the ``Table`` instance. Instead, you can just document that you need to call the ``describe()`` method to actually update the Table. But I do like the idea of removing that if statement in the ``describe()`` call to account for the ability to now create and delete gsi's for tables.

pauloubuntu:
@kyleknap I just removed that if statement and added a few notes about the need to call ```describe``` to update the ```Table``` instance. 

Apart from the fact we are not going to deprecate ```update``` due to a couple of reasons, I guess we've solved every point discussed. You may now merge this pull request if you want.

@kyleknap , @xethorn , @danielgtaylor 
I want to thank you for the time you've spent to help me and exposing your opinions. I've undoubtedly learnt a lot with this PR.

xethorn:
De nada
--
Sent from myMail app for Android
Thursday, 05 February 2015, 08:34PM -05:00 from Paulo Miguel Almeida <notifications@github.com>:
@kyleknap I just removed that if statement and added a few notes about the need to call describe to update the Table instance.
Apart from the fact we are not going to deprecate update due to a couple of reasons, I guess we've solved every point discussed. You may now merge this pull request if you want.
@kyleknap ,  @xethorn ,  @danielgtaylor
I want to thank you for the time you've spent to help me and exposing your opinions. I've undoubtedly learnt a lot with this PR.

kyleknap:
Awesome! Thanks for all of the work everyone put into this PR. Merging

PR #2926, opened 2015-02-02
Fix import for boto.cloudhsm.layer1.CloudHSMConnection.

abutcher:
Just started using boto.cloudhsm.connect_to_region() and found this.


kyleknap:
Thanks for the catch! Merging.

PR #2927, opened 2015-02-02
Fix region bug

danielgtaylor:
AWS Data Pipeline is not actually available in `eu-central-1` according to the [regions and endpoints](http://docs.aws.amazon.com/general/latest/gr/rande.html#datapipeline_region) documentation. This just removes the line that was mistakenly added.

cc @jamesls @kyleknap


kyleknap:
:ship:

Issue #2617, opened 2014-09-16
Boto config file parsing error


tarekziade:
Bug on Python 3.4 if I set the is_secure flag in [Boto]

```
Traceback (most recent call last):
 [...]
  File "/Users/tarek/Dev/github.com/loads-broker/lib/python3.4/site-packages/boto-2.32.1-py3.4.egg/boto/ec2/__init__.py", line 66, in connect_to_region
    return region.connect(**kw_params)
  File "/Users/tarek/Dev/github.com/loads-broker/lib/python3.4/site-packages/boto-2.32.1-py3.4.egg/boto/regioninfo.py", line 187, in connect
    return self.connection_cls(region=self, **kw_params)
  File "/Users/tarek/Dev/github.com/loads-broker/lib/python3.4/site-packages/boto-2.32.1-py3.4.egg/boto/ec2/connection.py", line 103, in __init__
    profile_name=profile_name)
  File "/Users/tarek/Dev/github.com/loads-broker/lib/python3.4/site-packages/boto-2.32.1-py3.4.egg/boto/connection.py", line 1081, in __init__
    profile_name=profile_name)
  File "/Users/tarek/Dev/github.com/loads-broker/lib/python3.4/site-packages/boto-2.32.1-py3.4.egg/boto/connection.py", line 478, in __init__
    is_secure = config.getboolean('Boto', 'is_secure')
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/configparser.py", line 798, in getboolean
    def getboolean(self, section, option, *, raw=False, vars=None,
  File "/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/configparser.py", line 776, in _get
    return conv(self.get(section, option, **kwargs))
TypeError: get() got an unexpected keyword argument 'raw'
```

tarekziade:
Ok so here's what I have to do until it gets fixed

```python
try:
    from boto.pyami.config import Config, ConfigParser

    def get(self, section, name, default=None, **kw):
        try:
            val = ConfigParser.get(self, section, name, **kw)
        except:
            val = default
        return val

    Config.get = get

except ImportError:
    pass
```

tarekziade:
poke @alfredodeza :)

jseabold:
This is all a bit silly. There are still a lot of projects out there that use `boto` and the blanket try/except swallowing a real error here makes this pretty hard to track down.

jamesls:
[...]

PR #2929, opened 2015-02-03
Python 3.4 compatibility on `is_secure` in `Boto` section of `boto.cfg`

grahamc:
This change is [Reviewable](https://reviewable.io/reviews/boto/boto/2929)


danielgtaylor:
This is related to #2617. This doesn't seem to fix the concerns in that issue.

tarekziade:
what do we need to do to make this happen ?

tarekziade:
I think the change should be done in boto.pyami.config.Config.get()  by adding **kw and passing them along. That will work for all py versions - and make sure all calls to getboolean() work

PR #2930, opened 2015-02-03
bSwitch to pyOpenSSL library

branto1:
python-rsa library exhibits several security flaws, this patch drops
the python-rsa dependency and implements the same functionality with
pyOpenSSL library.

### DIFF
--- a/boto/cloudfront/distribution.py
+++ b/boto/cloudfront/distribution.py
@@ -653,12 +653,13 @@ class Distribution(object):
         Requires the rsa library be installed.
         """
         try:
-            import rsa
+            import OpenSSL.crypto as ssl
         except ImportError:
-            raise NotImplementedError("Boto depends on the python rsa "
+            raise NotImplementedError("Boto depends on the pyOpenSSL"
                                       "library to generate signed URLs for "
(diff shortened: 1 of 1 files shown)


branto1:
Note: debian-based distros know this under the name python-openssl.

danielgtaylor:
@branto1 thanks for the pull request and trying to get this updated. It seems that PyOpenSSL actually depends on the [cryptography](https://github.com/pyca/cryptography) module:

https://github.com/pyca/pyopenssl/blob/master/setup.py#L37

So if we switch from `python-rsa` I think we'd just go with `cryptography` directly, unless there is a good reason not to. We are still deciding which way to move forward with this issue.

alex:
This has teh advantage that pyOpenSSL is packaged in older versions of debian, etc. and folks can still use that version.

branto1:
Yep, this should also work in fedora 20, rhel 7 and some other systems where there is no cryptography module.

kevgliss:
Any update on which way this is going to go? I will be working on a pull request adding signature verification to SNS messages.

I have no problem with either, but would lean towards cryptography if given a choice.

PR #2932, opened 2015-02-04
Ensure Content-Length header is a str

kevinschaul:
Ensure Content-Length header is a str. This is more explicit than relying on other code cast the value later.

This skirts an issue with httplib in Python 2.7.0, which assumes headers are already of type str. Though newer versions of httplib cast headers to str values, it seems more explicit for boto to do the casting.


danielgtaylor:
Thanks for the fix! :+1:

PR #2933, opened 2015-02-04
CORS add_rule(), copy/paste error preventing allowed_header check.

nordstromj:
S3 CORS clearly isn't being used very much. Simple copy paste error in cors.py "add_rule()" when passing a single "allowed header" as a string. allowed_origin variable was incorrectly being used twice when the second type check should have been allowed_header (similar to expose_header check). All other tags are handled correctly.


danielgtaylor:
Thanks for the pull request! Seems we don't have a test covering this use case. Would you mind adding a unit test for this?

nordstromj:
@danielgtaylor just added new test to cover this issue. Thanks!

Issue #2599, opened 2014-09-09
Incorrect parameters being passed to boto.manage.cmdshell.SSHClient


nathan-osman:
In `sshclient_from_instance()` ([boto/manage/cmdshell.py:406](https://github.com/boto/boto/blob/develop/boto/manage/cmdshell.py#L406)):

```
s = FakeServer(instance, ssh_key_file)
return SSHClient(s, host_key_file, user_name, ssh_pwd)
```

The `ssh_pwd` parameter is passed as the fourth parameter to the `SSHClient` constructor. However, the signature for the constructor ([boto/manage/cmdshell.py:51](https://github.com/boto/boto/blob/develop/boto/manage/cmdshell.py#L51)) is as follows:

```
def __init__(self, server,
             host_key_file='~/.ssh/known_hosts',
             uname='root', timeout=None, ssh_pwd=None):
```

The fourth parameter is `uname`, not `ssh_pwd`. This makes it impossible to use `sshclient_from_instance()` to connect to an instance using an encrypted private key. An exception is raised when this is attempted:

```
  ...
  File "/usr/local/lib/python2.7/dist-packages/boto/manage/cmdshell.py", line 411, in sshclient_from_instance
    return SSHClient(s, host_key_file, user_name, ssh_pwd)
  File "/usr/local/lib/python2.7/dist-packages/boto/manage/cmdshell.py", line 61, in __init__
    password=ssh_pwd)
  File "/usr/lib/python2.7/dist-packages/paramiko/pkey.py", line 198, in from_private_key_file
    key = cls(filename=filename, password=password)
  File "/usr/lib/python2.7/dist-packages/paramiko/rsakey.py", line 51, in __init__
    self._from_private_key_file(filename, password)
  File "/usr/lib/python2.7/dist-packages/paramiko/rsakey.py", line 163, in _from_private_key_file
    data = self._read_private_key_file('RSA', filename, password)
  File "/usr/lib/python2.7/dist-packages/paramiko/pkey.py", line 280, in _read_private_key_file
    data = self._read_private_key(tag, f, password)
  File "/usr/lib/python2.7/dist-packages/paramiko/pkey.py", line 323, in _read_private_key
    raise PasswordRequiredException('Private key file is encrypted')
paramiko.PasswordRequiredException: Private key file is encrypted
```

The fix for this is relatively simple. Change ([boto/manage/cmdshell.py:407](https://github.com/boto/boto/blob/develop/boto/manage/cmdshell.py#L407)) to:

[...]

PR #2934, opened 2015-02-05
Correct parameters for SSHClient

joerayme:
Pass the correct keyword parameters to SSHClient from sshclient_from_instance - this fixes issue #2599


danielgtaylor:
Thanks for the pull request. Would it be possible to add a unit test that calls this method to ensure it works as expected should the code change in the future? It seems that a lack of a test allowed this to happen.

joerayme:
Wrote a simple unit test as requested. Let me know if you think it needs to do a bit more!

PR #2935, opened 2015-02-05
Simplify metadata update

gregr:
This issue seems to be a thorn in the spine of many boto users, as seen by the number of questions asked about this topic on various sites.  Worse, these questions are usually answered incorrectly at worst and insufficiently at best.

Let me know if any changes need to be made for this PR to be acceptable.


boto/s3/key.py
```diff
@@ -545,6 +545,31 @@ def delete(self, headers=None):
         return self.bucket.delete_key(self.name, version_id=self.version_id,
                                       headers=headers)
 
+    def user_settable_metadata():
```
danielgtaylor: Missing `self` parameter.

danielgtaylor:
Overall I think this looks good, but before it can be merged you'll need to add some unit tests to cover the new methods. It might also be worth updating the S3 documentation/tutorial since this is such a major pain point for others. Thanks for the pull request!

gregr:
@danielgtaylor Added tests and a tutorial example.  Let me know if you need anything else.

gregr:
@danielgtaylor thoughts?

PR #2936, opened 2015-02-05
Add support for splitting/merging of AWS reserved instances.

jbylund:
I'd like to be able to split and modify reserved instances.  It's necessary to put InstanceType in the request to AWS when doing so (ref: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ReservedInstancesConfiguration.html).  This patch is working for me locally.


danielgtaylor:
Thanks for the pull request. This looks great, but could you please add a unit test to cover this new parameter? There should be plenty of existing tests in `tests/unit/ec2/...` that can be used as examples.

Answashe:
We have run into this issue as well. Any chance of getting this merged before the next Boto release?

jbylund:
Apparently merged as https://github.com/boto/boto/pull/3085

PR #2937, opened 2015-02-05
Correct docs around overriding SGs on ELBs

thieman:
This commit corrects a critical documentation error where the
`LoadBalancer.apply_security_groups` and
`ELBConnection.apply_security_groups_to_lb` methods incorrectly
state that they add security groups to a load balancer. These
methods are not additive, rather they override any currently
applied security groups with the security groups passed to them.

This is potentially catastrophic from a user's perspective as
a destructive method is being presented as non-destructive.

The AWS documentation for the ApplySecurityGroupsToLoadBalancer
action may be found here:
http://docs.aws.amazon.com/ElasticLoadBalancing/latest/APIReference/API_ApplySecurityGroupsToLoadBalancer.html

### DIFF
--- a/boto/ec2/elb/__init__.py
+++ b/boto/ec2/elb/__init__.py
@@ -695,9 +695,9 @@ class ELBConnection(AWSQueryConnection):
 
     def apply_security_groups_to_lb(self, name, security_groups):
         """
-        Applies security groups to the load balancer.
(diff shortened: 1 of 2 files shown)


danielgtaylor:
Thanks for the clarification! :+1:

PR #2948, opened 2015-02-09
switch to unquote_plus in bucketlistresultset

kouk:
Resubmitting #2654, rebased off current develop branch. Original description follows:

There is a bug in PR #2562, which I submitted. The URL quoting that S3 performs with `encoding_type="url"` necessitates the use of `unquote_plus` because they replace spaces with plus signs. This didn't break bucket listing completely, and in my case it only led to some keys missing from the listing, something that I didn't notice in my original testing. Sorry for this!


kouk:
(rebased again)

ifly4pie:
Hi, any chance this can be reviewed and merged soon? 

We are also frustrated by this bug.

Use case:
* Prefix path has a space in the name
* Bucket has more than 1000 keys in the prefix path. 
* Some key names have valid unicode characters that break XML, so encoding_type='url' is required.

`bucket.list(prefix='path/has a space/', encoding_type='url')`  is broken and only returns the first 1000 keys because of the plus sign in the marker's key name.

PR #2950, opened 2015-02-09
Added support for nonversion expiration and transition

mattford63:


anonymuse:
This would be great to have :+1:

danielgtaylor:
@mattford63 this looks generally good, but to be accepted it will need to add some tests to cover the new features. Thanks! :+1:

mattford63:
Hi, I added the tests, and fixed a couple of issues they brought up.

mattford63:
Hi, is there anything more I can do to help this along?

mattford63:
@danielgtaylor Hi :-) , is this to accepted or rejected?

kyleknap:
Sorry on the delay here. I am fine with this change but there is a few things that I would like to see updated:

* Could you rebase off of the HEAD of develop? Looks like this is not a clean merge as of now.
* There some Python styling issues. Could you make sure that all parameters and properties are snake_cased? Also can you make sure that the lines are shorter than 80 characters? You can check this one by running ``flake8``.

PR #2951, opened 2015-02-10
script "glacier upload ... " to return the ArchiveID

arterrey:
Helpful for logging and testing backups.


arterrey:
The ArchiveID is 138 characters long! Is that expected?

danielgtaylor:
This looks useful, thanks! :+1:

PR #2952, opened 2015-02-10
add some minor documentation for Route53 tutorial

jimmystewpot:
Leading users to get a light example of ResourceRecordSets which seems to be something missing based on some of the questions I had when I started working with the Route53 library.


danielgtaylor:
Thanks! :+1:

PR #2953, opened 2015-02-10
Fixes #2931 and documents args

### LINKED ISSUE #2931: health_check cannot run update method because missing self.connection
It seems healthcheck expects to have access to self.connection when running ```update()``` method, but this isn't an attribute.  The load balancer does have a connection though, so there is a work around for now...

```
>>> proxy_elb, web_elb = hvpc.get_load_balancers()
>>> proxy_elb
LoadBalancer:ficops01-proxy-elb
>>> hc = proxy_elb.health_check
>>> hc.healthy_threshold = 3
>>> hc.unhealthy_threshold = 3
>>> hc.interval = 20
```

This doesn't work:

```
>>> proxy_elb.health_check.update()
  File "<input>", line 1, in <module>
  File "/Users/russellballestrini/env/lib/python2.7/site-packages/boto/ec2/elb/healthcheck.py", line 83, in update
    new_hc = self.connection.configure_health_check(self.access_point,
AttributeError: 'HealthCheck' object has no attribute 'connection'
```

Work Around:
```
>>> proxy_elb.configure_health_check(hc)
HealthCheck:TCP:3128
```

russellballestrini:
Fixes #2931 and documents args

1. Unit tests pass 
1. Sphinx HTML docs build
1. The failing test case in issue #2931 works

	modified:   boto/ec2/elb/healthcheck.py

### DIFF
--- a/boto/ec2/elb/healthcheck.py
+++ b/boto/ec2/elb/healthcheck.py
@@ -24,28 +24,35 @@
 
 class HealthCheck(object):
     """
-    Represents an EC2 Access Point Health Check. See
-    :ref:`elb-configuring-a-health-check` for a walkthrough on configuring
-    load balancer health checks.
+    Represents an EC2 Access Point Health Check.
+    See :ref:`elb-configuring-a-health-check` for a walkthrough
+    on configuring load balancer health checks.
     """
-    def __init__(self, access_point=None, interval=30, target=None,
-                 healthy_threshold=3, timeout=5, unhealthy_threshold=5):
+    def __init__(self, load_balancer=None, timeout=5, target=None,
+                 interval=30, unhealthy_threshold=5, healthy_threshold=3):
         """
-        :ivar str access_point: The name of the load balancer this
-            health check is associated with.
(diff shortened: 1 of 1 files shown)


danielgtaylor:
@russellballestrini thanks for the pull request. Unfortunately this change is not backward compatible and would break existing users. A better solution would keep the argument names and their positions the same and just update the failing method to use the correct attribute name. What do you think?

russellballestrini:
bummer, I was really excited to fix the ordering of the arguments now, because the module was broken already anyways.  All the tests are passing, and I fixed up the docs.  Additionally there was exactly one place in the code that this class was instantiated, and it did so by sending self (in this case the load balancer object) as the first argument.  (https://github.com/boto/boto/blob/develop/boto/ec2/elb/loadbalancer.py#L135)

This change requires breaking existing functionality.  Because of this, I felt it was a natural time to fix the argument list and adjust the docs to explain the new way to create an object of this class.

russellballestrini:
created a PR that does what you described: #2957 

I'll let you pick which to merge and which to close

PR #2975, opened 2015-02-18
Fix eu central 1 region

hjacobs:
The endpoints for IAM and Route53 for the eu-central-1 (Frankfurt) region are missing in endpoints.json. The fix is trivial as these services are region-independent.


hjacobs:
Sorry, this PR contains too many commits ---- I created a new pull request with "develop" as base. Closing...

PR #2980, opened 2015-02-20
Maintain self.region when creating SWFBase objects

markdoliner:
Revision a79b8b7e0e0335d7c07cbdb804db53b793fe19c7 added support for
specifying the region when creating the Layer1 object.

This commit extends that capability a tiny bit: When a child object
of SWFBase instantiates another child object of SWFBase, it now passes
self.region through to the new instance.

Without this change, for example, this call:
    boto.swf.layer2.Domain(name='foo', region=region).workflows()
Returns a list of WorkflowType objects where region is not set, which
means the default region is used when calling e.g. WorkflowType.start(),
which is not what I would expect.


danielgtaylor:
The fix should now be in from #2985. If it's missing anything please feel free to comment.

PR #2981, opened 2015-02-20
EC2 connection filters can use tuple values

joncotton:
The first thing is simple, just grab the key and value from the dict at the same time.

Second change allows values to be given as tuples and not just lists. Solves this case:
```
>>> connection.get_only_instances(filters={"tag:Name": ["example", "test"]})
[Instance:i-b67dc459, Instance:i-5b9d20b4]
>>> connection.get_only_instances(filters={"tag:Name": ("example", "test")})
[]
```

### DIFF
--- a/boto/ec2/connection.py
+++ b/boto/ec2/connection.py
@@ -126,12 +126,11 @@ class EC2Connection(AWSQueryConnection):
             filters = dict(filters)
 
         i = 1
-        for name in filters:
+        for name, value in six.iteritems(filters):
             aws_name = name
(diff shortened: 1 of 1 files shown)


joncotton:
Could you provide some guidance on where the tests should go? There are no current tests for `build_filter_params()`. Should I create something like this in `tests/unit/ec2/test_connection.py`?

```
class TestBuildFilterParams(TestEC2ConnectionBase):
    def test_1(self):
        ...
    def test_2(self):
        ...
```

PR #2985, opened 2015-02-25
fix for regions in SWF layer2 - including unit tests

laflechejonathan-amzn:
Similarly to a few existing pull requests:

#2606
#2980

I've had to work around this bug in boto where the region is not propagated to other layer2 objects on calls to:

* workflows()
* activities()
* executions()


laflechejonathan-amzn:
Note - Travis CI error'ed out - I have no way to restart the build.

danielgtaylor:
Looks like [Travis CI is down](http://www.traviscistatus.com/incidents/bc3jgcrjs7gt). I've run the builds locally and everything seems to work. Thanks! :+1:

Issue #1505, opened 2013-05-20
RDSConnection.get_all_dbparameters always returns one less item


rbondoc-palominodb:
get_all_dbparameters is always returning one less parameter.

For example, I have this code.

``` python
pg = conn.get_all_dbparameters(pg.name)
```

The expected number of parameters returned is 100.

However, the command below only shows 99.

``` python
print len(pg.keys())
```

I already tried changing the max_records parameter but it always return one item less than the expected count.

Digging deeper, found at the bottom is the response from AWS API. It returned all 100 items. With this, maybe there's something wrong with the xml parser.

Moreover, I think this line produces the problem. I'm not sure if there is some kind of race condition happening. Executing the line twice seems to fix the problem.

``` python
xml.sax.parseString(body, h)
```

XML Response:

``` xml
DescribeDBParametersResponse xmlns="http://rds.amazonaws.com/doc/2012-09-17/">
  <DescribeDBParametersResult>
    <Marker>bG9ja193YWl0X3RpbWVvdXQ=</Marker>
    <Parameters>
      <Parameter>
        <DataType>boolean</DataType>
        <Source>engine-default</Source>
        <IsModifiable>false</IsModifiable>
        <Description>Controls whether user-defined functions that have only an xxx symbol for the main function can be loaded</Description>
        <ApplyType>static</ApplyType>
        <AllowedValues>0,1</AllowedValues>
        <ParameterName>allow-suspicious-udfs</ParameterName>
      </Parameter>
      <Parameter>
        <DataType>integer</DataType>
        <Source>engine-default</Source>
        <IsModifiable>true</IsModifiable>
        <Description>Intended for use with master-to-master replication, and can be used to control the operation of AUTO_INCREMENT columns</Description>
        <ApplyType>dynamic</ApplyType>
        <AllowedValues>1-65535</AllowedValues>
        <ParameterName>auto_increment_increment</ParameterName>
      </Parameter>
      <Parameter>
        <DataType>integer</DataType>
        <Source>engine-default</Source>
        <IsModifiable>true</IsModifiable>
        <Description>Determines the starting point for the AUTO_INCREMENT column value</Description>
        <ApplyType>dynamic</ApplyType>
        <AllowedValues>1-65535</AllowedValues>
        <ParameterName>auto_increment_offset</ParameterName>
      </Parameter>
      <Parameter>
        <DataType>boolean</DataType>
        <Source>engine-default</Source>
        <IsModifiable>true</IsModifiable>
[...]

PR #2991, opened 2015-02-26
Fix in-place dictionary key deletion.

contracode:
Fixes KeyError raised in boto.rds.create_dbinstance() by loop that deletes dictionary keys that the loop itself references.


contracode:
Added additional commit to fix issue #1505.

PR #3001, opened 2015-03-04
enable .egg/zipfiles, use pkg_resources

nyetsche:
by using a .egg file directly (i.e,. sys.path.insert(0,"boto-2.36.0-py2.7.egg") we
lose the os.path.* making any sense; pkg_resources was designed to fix this
https://pythonhosted.org/setuptools/pkg_resources.html#resource-extraction


nyetsche:
I believe this also solves https://github.com/boto/boto/issues/2872 and https://github.com/boto/boto/pull/2804 although I haven't checked. It works in my following use case:

`cd $BOTO-SOURCE && python ./setup.py bdist_egg && mv dist/*.egg $PROJECT && cd $PROJECT`

Followed by in $PROJECT:

```
import sys
sys.path.insert(0,"boto-$EGG.egg")
import boto
```

That is, I want a self-contained `.egg` I can source from. Without the change, I get errors like:

```
IOError: [Errno 20] Not a directory: 'boto-$VERSION.egg/boto/endpoints.json'
```

PR #3003, opened 2015-03-04
I believe bucket versioning is no longer in beta.  Updating docstring

shaunbrady:
S3 bucket versioning has come out of beta: http://aws.amazon.com/about-aws/whats-new/2010/02/08/versioning-feature-for-amazon-s3-now-available/.  The comment hung me up a little bit thinking it was the boto code it was referring to.  If the comment was referring to the status of the AWS feature, it is no longer true.

Thanks!

SB

### DIFF
--- a/boto/s3/bucket.py
+++ b/boto/s3/bucket.py
@@ -1262,8 +1262,6 @@ class Bucket(object):
         """
         Configure versioning for this bucket.
 
-        ..note:: This feature is currently in beta.
-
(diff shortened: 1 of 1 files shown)


shaunbrady:
I noticed that this branch has failed, but I believe this was due to Travis timing out.  Can someone from the project nudge this build or give me a suggestion on how to do it myself.  Thanks!

SB

Issue #3005, opened 2015-03-05
AWS Config service module does not appear to support the connect_to_region() method


georgej-github:
The connect_to_region() does not seem to be available for / work with the connect_to_region() method.  E.g.:

>>> import boto.configservice
>>> conn = boto.configservice.connect_to_region("us-west-2")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/boto-2.36.0-py2.7.egg/boto/configservice/__init__.py", line 38, in connect_to_region
    for region in regions():
  File "/usr/local/lib/python2.7/dist-packages/boto-2.36.0-py2.7.egg/boto/configservice/__init__.py", line 33, in regions
    from boto.kms.layer1 import ConfigServiceConnection
ImportError: cannot import name ConfigServiceConnection

Is it a bug? Or is the incorrect service module being used for AWS Config in the above code?   Please advise, thanks.

PR #3009, opened 2015-03-05
Add more regions for configservice

kyleknap:
Fixes https://github.com/boto/boto/issues/3005

cc @jamesls @danielgtaylor

### DIFF
--- a/boto/configservice/__init__.py
+++ b/boto/configservice/__init__.py
@@ -30,7 +30,7 @@ def regions():
     :rtype: list
     :return: A list of :class:`boto.regioninfo.RegionInfo`
     """
-    from boto.kms.layer1 import ConfigServiceConnection
+    from boto.configservice.layer1 import ConfigServiceConnection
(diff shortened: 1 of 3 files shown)


boto/configservice/__init__.py
```diff
@@ -30,7 +30,7 @@ def regions():
     :rtype: list
     :return: A list of :class:`boto.regioninfo.RegionInfo`
     """
-    from boto.kms.layer1 import ConfigServiceConnection
```
jamesls: Interesting, was this completely broken before?

jamesls:
:shipit:

boto/configservice/__init__.py
kyleknap: If you tried ``connect_to_region()``, I believe so. But since there was only one region ``us-east-1`` you could work around it with ``connect_configservice()``

StivOstenberg:
I am failing trying to call configservice.get_regions:
AttributeError: 'module' object has no attribute 'configservice'

PR #3014, opened 2015-03-06
AWS S3 Signature V4 - aws-chunked and other fixes

tpodowd:
This set of patches extends Boto's S3 Signature V4 support to handle 'aws-chunked' content-encoding. 

There are 3 ways to sign a payload using sigv4.
1. sign the entire payload (this is the existing implementation)
2. sign the payload in chunks (aws-chunked)
3. same as 2 but use HTTP chunked tranfer encoding.

These patches add 2 and 3. It also makes 2 the default signing way so as to avoid reading in a large file and first creating the hash before finally uploading. Using 2, each chunk is hashed and signed, so the body is effectively read once and sent faster. The Java SDK also seems to use this as their default. For sigv4 requests I've also removed the content-md5 hash calculation and header as it is no longer necessary when using sigv4.

Details are in each commit. We are using this feature to test our Cloudian S3 object store and I've also tested quite a lot against Amazon to ensure I haven't broken anything.

Would love this merged sooner than later so I don't have to maintain it as code changes as its quite big. I'll add test results below.


tpodowd:
Here is the S3 test run:

    [tpodowd@bertha tests (s3_sigv4_chunking)]$ ./test.py -t s3
    nose command: ./test.py -a !notdefault,s3
    ................................................................
    Ran 64 tests in 246.357s

    OK

tpodowd:
@danielgtaylor, @garnaat - I guess you guys should review this. Let me know if you have questions or concerns and I'll work on them. It was quite hard to work the chunking feature in but I tried to keep it as clean as possible without changing too much.

@mfschwartz - I've tested against the GS stuff also and everything looks good. One thing to note is that I found (and fixed) a bug in the code regarding chunked transfer encoding. Basically the Boto code was incorrectly putting in an additional ';' character which was breaking the protocol. AWS was actually rejecting the connection because of that but Google seemed to have been accepting it. Anyway, removed the ';' now and its correct according to the RFC. I'll include the output of the gs test as soon as it finishes.

tpodowd:
gs test finally finished. Here is the result.

    [tpodowd@bertha tests (s3_sigv4_chunking)]$ ./test.py -t gs
    nose command: ./test.py -a !notdefault,gs
    ......................................................................
    Ran 70 tests in 1190.342s

    OK

tpodowd:
Hmm. Seems Travis CI Build thingie has some error though. I'll investigate this.

tpodowd:
Ok. still quite a few things broken. I've managed to repeat the test cases in my environment with a similar output so I'll work on them one by one. Might be a weekend between now and then though...

tpodowd:
Right. I added another commit which hopefully addresses the errors. When I run the "test.py default" in my environment I get the same "FAILED (SKIP=6, errors=6, failures=4)" output for both the develop branch and this branch. Let's see what Travis says.

tpodowd:
Travis CI is still saying failed but clicking on the details all the tests are working except those for Python 3.4 which looks like it is hanging. Looking at other pull requests many of them have the same issue so I suspect this is not a problem with this pull request but rather something more general. Previous issues have been fixed and ready for review again.

jamesls:
Could you rebase against develop and re push you branch?  We had to fix an issue in our test deps that was causing the travis CI builds to hang.  Thanks.

tpodowd:
Hi @jamesls , Thanks for the heads up. Ok, I rebased my patches against develop again. Looks like Travis CI says all is well this time.

In addition here is a newly run s3 test output:

    [tpodowd@bertha tests (s3_sigv4_chunking)]$ ./test.py -t s3
    nose command: ./test.py -a !notdefault,s3
    ................................................................
    Ran 64 tests in 244.209s

    OK

And I'm waiting on the GS test currently. It takes about 20 minutes to run.

tpodowd:
Ok, here is the GS test output.

    [tpodowd@bertha tests (s3_sigv4_chunking)]$ ./test.py -t gs
    nose command: ./test.py -a !notdefault,gs
    ......................................................................
    Ran 70 tests in 1128.741s
    
    OK

tpodowd:
I found another V4 issue with double ports in the host header. Adding the patch here. This doesn't effect port 80 or port 443, only non-standard ports.

Additionally, this change only effects sigv4 so no need to test gs this time.

    [tpodowd@bertha tests (s3_sigv4_chunking)]$ ./test.py -t s3
    nose command: ./test.py -a !notdefault,s3   
    ................................................................
    Ran 64 tests in 239.752s
    
    OK

tpodowd:
I added one more fix. Signature V4 doesn't like it when you have multiple headers of the same name. I squashed content-encoding in particular down to one header when using v4 and chunking.

    $ ./test.py -t s3
    nose command: ./test.py -a !notdefault,s3
    ................................................................
    Ran 64 tests in 259.788s
    
    OK

tpodowd:
I have just rebased this set of patches against the latest develop branch. After more experience using the Java SDK and these patches, I have also updated how the region name is handled in the scope.

The new behavior is:

     1. Lookup the S3 endpoint in User endpoints.json file
        pointed at by the Boto.endpoints_path variable and
        return the matching region.
     2. Lookup the S3 endpoint in System endpoints.json file
        and return the matching region
     3. Parse the endpoint and pull out the region name from
        it if possible. The following will be matched:
            s3.<region>.xyz
            <bucket>.s3.<region>.xyz
            s3.<region>.<domain>
            <bucket>.s3.<region>.<domain>
            s3-<region>.<domain>
            <bucket>.s3-<region>.<domain>
     4. If the region cannot be matched from above, an exception
        will be thrown and the message asking the user to setup
        an endpoints.json file as per rule 1 will be logged.
     5. Should rules 2 to 3 expand to a bad region, the user
        can also override this using rule 1.

The tests pass also.

    $ ./test.py -t s3
    nose command: ./test.py -a !notdefault,s3
    ................................................................
    Ran 64 tests in 276.715s
    
    OK

tpodowd:
@danielgtaylor - any chance you can review this? Or suggest who else will.

We are now using this patch and a few others that I have filed against both AWS and Cloudian signature V4 on all our testing so if there will be updates or changes, I'd love to get them in sooner than later.

jmunhoz:
@jamesls is this patchset planned to be merged in some estimated date? I watched several old PRs fixing the double port bug with boto2 and they looks forgotten. Should we contact some concrete boto2 fellow to see those patches upstream?

tpodowd:
I notice that this branch has a conflict now. I'll try to grab some time to rebase and retest it again.

tpodowd:
Hi @jamesls - just rebased this pull request on latest develop. S3 tests look good as before.

tpodowd:
Hi @jamesls, @jmunhoz - I added another patch to this patch set which adds support for SigV4 unsigned payloads. The extra feature was added by @mozawa on top of my existing patches and I additionally added some code docs and updated the test cases to test the feature.

Would love feedback on this as we have been using this in house to test our Cloudian SigV4 feature since last year and have incorporated it into all out tools. I feel it is a very useful set of features to get into Boto and will likely also help the Ceph S3 Test guys.

tpodowd:
@jamesis - fixed another bug that crept in when I rebased onto 2.42. Added a unit test for it also.

PR #3018, opened 2015-03-08
force float division while calculating chunk_count

farthVader:
the math.ceil is rendered useless since integer division is performed (source_size, chunk_size are integers).


danielgtaylor:
Oops, you're right. Thanks for fixing the docs! FWIW, it is done correctly in `s3put`:

https://github.com/boto/boto/blob/develop/bin/s3put#L210

Issue #2605, opened 2014-09-10
CloudSearch2 SearchResult.facet disregarding item order


nealclark:
The facets in the SearchResults.facets object are loaded into a python dictionary which discards the ordering provided by Amazon CloudSearch. 

``` python
facet = {
    'facet_name': {
        'sort': 'bucket'
    },
}

# execute search
searchservice = domain.get_search_service()
results = searchservice.search(
    q='matchall',
    parser='structured',
    return_fields=['id'],
    size=1,
    facet=facet)

print results.facets
```

Outputs:

``` python
{
    u'facet_name': {
        u'c': 1,
        u'a': 4,
        u'b': 2,
    }
}
```

The same is true if you were to try to sort by 'count'.

PR #3022, opened 2015-03-10
Fix Cloudsearch facets sort order.

tell-k:
Facet of Cloudsearch returns the results in a state where the sort.
However, because the code of the bot is utilized dict, and thus to ignore the results of the sorting.

Therefore, I have modified the code to take advantage of OrderedDict.
And fixed tiny bug :)

Related: #2605

### DIFF
--- a/boto/cloudsearch/search.py
+++ b/boto/cloudsearch/search.py
@@ -22,8 +22,13 @@
 # IN THE SOFTWARE.
 #
 from math import ceil
-from boto.compat import json, map, six
+from boto.compat import json, six
 import requests
+
+try:
+    from collections import OrderedDict
+except ImportError:
+    from ordereddict import OrderedDict
 
 
 class SearchServiceException(Exception):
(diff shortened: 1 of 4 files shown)


boto/cloudsearch/search.py
```diff
@@ -22,9 +22,14 @@
...
 #
 from math import ceil
-from boto.compat import json, map, six
+from boto.compat import json, six
 import requests
 
+try:
+    from collections import OrderedDict
+except ImportError:
+    from ordereddict import OrderedDict
```
danielgtaylor: Can we move this logic into the `boto.compat` module? Then it's just `from boto.compat import OrderedDict` instead.

Also, this is introducing a new dependency. What about users on 2.6.x that don't have the OrderedDict module installed? We may need to fall back to an ordinary `dict` so that existing code in the wild does not break.

boto/cloudsearch/search.py
```diff
@@ -101,7 +109,7 @@ def __init__(self, q=None, bq=None, rank=None,
     def update_size(self, new_size):
         self.size = new_size
         self.real_size = Query.RESULTS_PER_PAGE if (self.size >
-            Query.RESULTS_PER_PAGE or self.size == 0) else self.size
+                                                    Query.RESULTS_PER_PAGE or self.size == 0) else self.size
```
danielgtaylor: PEP8 line length

danielgtaylor:
Thanks for the pull request. This seems like a useful addition, I just want to make sure we don't break anyone. Nice job on the unit tests :+1:

natecode:
@danielgtaylor Now that it has tests, what is blocking this being merged?

PR #3023, opened 2015-03-10
EC2Connection should not ignore the host parameter

haridsv:
I noticed that the `host` parameter is being completely ignored in favor of what is coming from `config`.

### DIFF
--- a/boto/ec2/connection.py
+++ b/boto/ec2/connection.py
@@ -96,7 +96,7 @@ class EC2Connection(AWSQueryConnection):
                                             aws_secret_access_key,
                                             is_secure, port, proxy, proxy_port,
                                             proxy_user, proxy_pass,
-                                            self.region.endpoint, debug,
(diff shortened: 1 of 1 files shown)


danielgtaylor:
Would it be possible to add a unit test which shows the use case this is trying to fix? Thanks!

haridsv:
Should be possible, I will take a look at the test setup.

haridsv:
I added a unit test and verified that one of the tests fails without the fix. Apologies for not adding the test in the first place, didn't read through the CONTRIBUTING guidelines :)

haridsv:
@danielgtaylor is the unit test satisfactory?

haridsv:
For the sake of completeness, I am fixing other connection classes that have the same problem, so please hold on.

haridsv:
@danielgtaylor I have now searched and updated all the various connection classes to support the host parameter uniformly and also added unittests for each one of those. Please review the changes.

Issue #2296, opened 2014-05-29
CloudFormation UpdateStack call does not support UsePreviousValue for parameters


sigmaris:
There is an API parameter 'UsePreviousValue' that can be passed in for stack Parameters when performing a Cloudformation stack update, but there is no way at the moment to set this parameter when updating a stack using boto.

http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_Parameter.html

aronkyle-lightspeedretail:
I have a pull request up to address this issue (posted on oct 31st). Could someone get me some feedback on this?

tomyan:
Any update on this - would be really useful?

aronkyle-lightspeedretail:
@tomyan 
I not sure the best way to get the maintainers attention. This pull request has been sitting here for months now.

jantman:
This is a blocker for my use of boto, as we randomly generate database passwords when we spin up a stack. Any chance of getting eyes on this?

joekiller:
+1 to add this

tomyan:
We worked around this by using the same private methods that update_stack uses. We first call `boto.cloudformation._build_create_or_update_params`, then take the params that result and manipulate that so that the parameters are left alone (replace once ending ".ParameterValue" with ones ending ".UsePreviousValue" with value "true") before passing the whole thing to `boto.cloudformation._do_request`. We then mitigate the risk of these apis changing by pinning to a particular version of boto. Pretty ugly though - would love to see this pull request applied so we can remove this horrible hack.

Tom

joekiller:
Yeah I just monkey patched this on top of the existing method and moved along.

```python
__cfn_conn = None

def connect_cloudformation():
    from boto.cloudformation.connection import CloudFormationConnection

    def _build_create_or_update_params(self, stack_name, template_body,
                                       template_url, parameters, disable_rollback, timeout_in_minutes,
                                       notification_arns, capabilities, on_failure, stack_policy_body,
                                       stack_policy_url, tags, stack_policy_during_update_body=None,
                                       stack_policy_during_update_url=None):
        params = {'ContentType': "JSON", 'StackName': stack_name,
                  'DisableRollback': self.encode_bool(disable_rollback)}
        if template_body:
            params['TemplateBody'] = template_body
        if template_url:
            params['TemplateURL'] = template_url
        if template_body and template_url:
            boto.log.warning("If both TemplateBody and TemplateURL are"
                             " specified, only TemplateBody will be honored by the API")
[...]

PR #3029, opened 2015-03-12
Issue 2296. Adding support for UsePreviousValue.

aronkyle-lightspeedretail:
Adding support for cloudformation UsePreviousValue parameter.

In response to issue #2296.


tests/integration/cloudformation/test_connection.py
```diff
@@ -114,9 +131,37 @@ def test_large_template_stack_size(self):
...
+             parameters=[('Parameter1', '', True),
+                         ('Parameter2', 'updated_value')])
+        
+        stacks = self.connection.describe_stacks(self.stack_name)
+        stack = stacks[0]
+        params = [(p.key, p.value) for p in stacks[0].parameters]
+        self.assertEquals([('Parameter1', 'initial_value'),
+                           ('Parameter2', 'updated_value')], params)
+        
+        for _ in range(30):
```
kyleknap: What is the reason for sleeping again when nothing else is happening after other than cleaning up the stack by deleting it?

tests/unit/cloudformation/test_connection.py
```diff
@@ -149,6 +150,11 @@ def test_update_stack_all_args(self):
             'NotificationARNs.member.2': 'arn:notify2',
             'Parameters.member.1.ParameterKey': 'KeyName',
             'Parameters.member.1.ParameterValue': 'myKeyName',
+            'Parameters.member.2.ParameterKey': 'KeyName2',
+            'Parameters.member.2.ParameterValue': '',
```
kyleknap: You may need to update the logic to exclude ``ParameterValue`` if ``UsePreviousValue`` is being used. It looks like it works according to the integration test, but I would like to stick with what the API references say, which is to exclude a parameter value if specify UsePreviousValue is true: http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_Parameter.html

kyleknap:
@aronkyle-lightspeedretail 
Thanks for the pull request! It looks good to me in general. I just had a couple comments that I would like you to address before I merge it.

tests/integration/cloudformation/test_connection.py
aronkyle-lightspeedretail: It's been quite a few months but I think it had to do with stack.update() responding immediately with a status of IN PROGRESS. So we need some kind of delay to keep the test going long enough for the update to actually take affect before attempting the update with the "previous value" in a second update. 

I'm open suggestions if there's a better way.

tests/unit/cloudformation/test_connection.py
aronkyle-lightspeedretail: Again, it's been a few months. But I don't think the API  will accept it otherwise.

tests/integration/cloudformation/test_connection.py
kyleknap: But isn't that what you do at this wait (which I am fine with): https://github.com/boto/boto/pull/3029/files#diff-e70f6fdc16bd22f32c84d3977d1d9f5cR142

I was just confused why you wait again after you test that the stack successfully has been updated such that it uses both the new and the previous values.

aronkyle-lightspeedretail: Oh I see. I think that is just to guarantee that cloud formation is in a "good" state when the test completes. e.g. a tear down could potentially fail if the stack was still updating.

If you want that out I can remove as soon as a find some time.

kyleknap: Cool. Yeah I would prefer if you remove it.

tests/unit/cloudformation/test_connection.py
aronkyle-lightspeedretail: I'll try and find some time to remove and test the above change. If I can update the logic I will. Otherwise we might have to leave it as is.

kyleknap: Thanks! I think we will need to do the update to the logic.

I just tested the integration tests where you set the previous value to true:
https://github.com/boto/boto/pull/3029/files#diff-e70f6fdc16bd22f32c84d3977d1d9f5cR151

It looks like this passes because the value in the second element of the tuple is ``''``. If you put a value like ``'foo'`` in as the second element, you get the following error:
```
boto: ERROR: 400 Bad Request
boto: ERROR: {"Error":{"Code":"ValidationError","Message":"Invalid input for parameter key Parameter1. Cannot specify usePreviousValue as true and non empty value for a parameter","Type":"Sender"},"RequestId":"b6c9aeeb-cdab-11e4-9e3d-df34bd1ee4bf"}
```

Don't worry, the fix is easy to make the ``'foo'`` test case work.

Just make the logic in ``connection.py``:
```
params['Parameters.member.%d.ParameterKey' % (i + 1)] = key
if use_previous:
    params['Parameters.member.%d.UsePreviousValue' 
                 % (i + 1)] = self.encode_bool(use_previous)
else:
    params['Parameters.member.%d.ParameterValue' % (i + 1)] = value

```
I tested the integration test and that worked for me.

aronkyle-lightspeedretail:
@kyleknap 

I updated the call to omit the parameter value if user previous value is true.

I also removed the extra wait loop at the end of the integration test.

That said, if you run just the one test, the test passed, but an error is returned trying to delete the stack since it hasn't finished updating yet.

kyleknap:
@aronkyle-lightspeedretail 

Oh, if an error is returned when you try to delete the stack, then we should keep the extra wait loop. It is more important that resources are cleaned up. Thanks for hanging in there with me.

aronkyle-lightspeedretail:
@kyleknap The wait loop is back in. Integration test passes.

kyleknap:
Thanks for seeing the PR through! Merging.

aronkyle-lightspeedretail:
Cheers. ;)

PR #3031, opened 2015-03-13
Made 'Items' structured objects have members=True

rrealmuto:
for mws.connections.get_fulfillment_preview and create_fulfillment_order

Calls to these functions failed because the Items object needs the form Items.member.n but because members was not set to True in the decorator, it would just fail.

### DIFF
--- a/boto/mws/connection.py
+++ b/boto/mws/connection.py
@@ -644,7 +644,8 @@ class MWSConnection(AWSQueryConnection):
         return self._post_request(request, kw, response)
 
     @requires(['Address', 'Items'])
-    @structured_objects('Address', 'Items')
+    @structured_objects('Address')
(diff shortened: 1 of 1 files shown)


disruptek:
Thanks for the fix; LGTM but the maintainers are going to want a test case demonstration success (and failure) to prevent regression...

rrealmuto:
What do you guys want to see as a test case? Just some code demonstrating
the function and output showing failure before the fix and working
afterwards?

disruptek:
Yep, you can take a look at the existing tests to see some contrived examples.  Add your test there and push the commit so that we don't accidentally break these calls in the future.

disruptek:
Added to my branch; thank you.

PR #3052, opened 2015-03-19
Fix deserialization of a CloudFormation template's NoEcho value.

chipx86:
When validating a CloudFormation template, the `NoEcho` value is converted
to a boolean. Since the incoming value is a string, and all non-empty
strings are considered true, the resulting `no_echo` value always ends
up being `True`, even if the value is `"false"`.

To fix this, the code has been updated to compare directly against
`"true"`.


chipx86:
Hey, just checking on the status of this. Still hitting this issue in 2.38.

Issue #2235, opened 2014-04-24
boto.ec2.connection.get_image throws exception when AMI ID is invalid.


maitreya1975:
The documentation (http://docs.pythonboto.org/en/latest/ref/ec2.html#boto.ec2.connection.EC2Connection.get_image) states that the method get_image 
Returns:  
The EC2 Image specified or None if the image is not found.

However, if the AMI ID is not found, it throws an exception instead of returning None. 

Sample Verbose output when invoking get_image with an invalid AMI ID: 

```
Traceback (most recent call last):
  File "getami.py", line 6, in <module>
    image = ec2.get_image('ami-c2492bf3')
  File "/usr/local/lib/python2.7/site-packages/boto/ec2/connection.py", line 260, in get_image
    return self.get_all_images(image_ids=[image_id], dry_run=dry_run)[0]
  File "/usr/local/lib/python2.7/site-packages/boto/ec2/connection.py", line 186, in get_all_images
    [('item', Image)], verb='POST')
  File "/usr/local/lib/python2.7/site-packages/boto/connection.py", line 1157, in get_list
    raise self.ResponseError(response.status, response.reason, body)
boto.exception.EC2ResponseError: EC2ResponseError: 400 Bad Request
<?xml version="1.0" encoding="UTF-8"?>
<Response><Errors><Error><Code>InvalidAMIID.NotFound</Code><Message>The image id '[ami-c2492bf3]' does not exist</Message></Error></Errors><RequestID>3dd34687-b22b-4dc9-96f3-3a283b1ecd1a</RequestID></Response>
```

Constantin07:
I'm facing the same issue and it happens sporadically.

```
    img = ec2.get_image(image_id)
File "/usr/lib/python2.7/site-packages/boto/ec2/connection.py", line 263, in get_image
    return self.get_all_images(image_ids=[image_id], dry_run=dry_run)[0]
File "/usr/lib/python2.7/site-packages/boto/ec2/connection.py", line 189, in get_all_images
    [('item', Image)], verb='POST')
File "/usr/lib/python2.7/site-packages/boto/connection.py", line 1186, in get_list
    raise self.ResponseError(response.status, response.reason, body)
    boto.exception.EC2ResponseError: EC2ResponseError: 400 Bad Request
    <?xml version="1.0" encoding="UTF-8"?>
[...]

PR #3060, opened 2015-03-25
Issue #2235. Catch EC2ResponseError in connection.get_image

mrdrozdov:
Inspired by the now closed:
https://github.com/firesh/boto/commit/9779706f628b5a2ef7c3a73da1a463b9019c1edd

### DIFF
--- a/boto/ec2/connection.py
+++ b/boto/ec2/connection.py
@@ -261,7 +261,7 @@ class EC2Connection(AWSQueryConnection):
         """
         try:
             return self.get_all_images(image_ids=[image_id], dry_run=dry_run)[0]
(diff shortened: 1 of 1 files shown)


mrdrozdov:
This is a small change to catch the EC2ResponseError that get's thrown here most of the time (None is meant to be returned).

@kyleknap could you take a look?

nchammas:
Is no image available the only possible cause of `EC2ResponseError` in this case?

mrdrozdov:
@nchammas I am not sure, but it certainly throws EC2ResponseError as well as IndexError. It's ambiguous whether EC2ResponseError should be bubbled up. Am closing this issue for the time being.

PR #3067, opened 2015-03-27
Fix _url_base64_encode of cloudfront in Python 3

sodastsai:


kyleknap:
Can you add tests for what you added?

PR #3068, opened 2015-03-27
correct return type for list_multipart_uploads

ilogik:
The docs show the wrong return type for boto.s3.bucket.list_multipart_uploads, namely BucketListResultSet instead of MultiPartUploadListResultSet.

http://boto.readthedocs.org/en/latest/ref/s3.html#boto.s3.bucket.Bucket.list_multipart_uploads


kyleknap:
Thanks! Merging.

PR #3074, opened 2015-04-01
Update AWS CloudTrail to the latest API

danielgtaylor:
This update adds new exceptions, removes deprecated optional parameters and adds the new `lookup_events` method.

**Note**: this includes a backward incompatible change that the service made. The `trail` parameter is no longer accepted and has been removed, which is why the test needed to be updated.

cc @jamesls @kyleknap


kyleknap:
LGTM :ship:

PR #3083, opened 2015-04-07
TCP Health Check and Tagging Capabilities

tzuryby:
Dears,

I needed to fix the code to enable TCP health check had to make <ResourcePath> an optional element.
Then, when creating health checks with the API, they are "name-less", and for this, I added tagging (add / remove) so as one set tag: {"Name": "my-hc-name"} - this becomes the Health-check name in the UI.


tzuryby:
Dears,

I just added a safer version, handling cases where params are not provided, available at https://github.com/tzuryby/boto/commit/37f29ff221308856898d52f3c66c52b8c6902fc3

kyleknap:
It would be good to get the tests to pass before we can look into the PR.

PR #3085, opened 2015-04-08
Fix issue with modify reserved instances

kyleknap:
Add ability to specify InstanceType and fixed another failing test.
Here is the API docs for the configuration structure: http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_ReservedInstancesConfiguration.html

cc @danielgtaylor @jamesls


jamesls:
:shipit:

PR #3086, opened 2015-04-09
Make the bundled copy of six a fallback

gholms:
As a package maintainer I need to [remove](https://fedoraproject.org/wiki/Packaging:No_Bundled_Libraries) bundled libraries and instead point things at the systemwide copies of them.  This patch makes boto attempt to use a systemwide copy of six first, and if it isn't there, it seamlessly uses the bundled one instead.

### DIFF
--- a/boto/compat.py
+++ b/boto/compat.py
@@ -46,15 +46,23 @@ except (AttributeError, ImportError):
     # This is probably running on App Engine.
     expanduser = (lambda x: x)
 
-from boto.vendored import six
-
-from boto.vendored.six import BytesIO, StringIO
-from boto.vendored.six.moves import filter, http_client, map, _thread, \
-                                    urllib, zip
-from boto.vendored.six.moves.queue import Queue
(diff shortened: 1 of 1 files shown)


kyleknap:
I do not know if we can take this PR. Mainly because the code may be relying on a version of six that is vendored in boto, and if users are using the wrong version of six this may cause boto to break. Is there a particular reason that you do not want to use the current vendored version of six or is it just is because it is a vendored package?

gholms:
It isn't specific to six; most Linux distros simply disallow library bundling altogether.  You're right to be concerned about having an appropriate version of six, though.  That is why this patch wraps _all_ of the import statements rather than just a single one.  If any of them fails it will fall back to the bundled copy.

An even better fix would be to simply declare a dependency on `six >= 1.7` in setup.py/requirements.txt/etc like boto does for all of the other libraries it depends on, but I assume you chose to treat this library differently for a reason.

gholms:
What test cases does this need that the existing suite does not already cover?

PR #3093, opened 2015-04-12
Round SQS float arguments up, not down, when an int is required

thieman:
Currently, parts of the SQS module related to delay and visibility timeouts cast their arguments to `int` within the SQS connection. A Python `int` cast rounds down, e.g. `int(3.99) == 3`. This causes unexpected behavior, as a developer who tries to delay an item for 3.99 seconds and receives a successful response surely will not expect that item to be made available in the queue after only 3 seconds.

This commit uses `math.ceil` to forcibly round up these arguments before casting them to `int` type. This incurs more delay than the developer may have expected, but the spirit of the constraints provided by SQS will remain true, e.g. "The item will not be made available before X seconds have passed," where X is whatever the developer passed in.

I would argue that the best behavior here would be to throw a `TypeError` if a `float` is passed to a function which expects an `int`, but this commit eschews that approach in favor of deviating as little as possible from current functionality.

### DIFF
--- a/boto/sqs/connection.py
+++ b/boto/sqs/connection.py
@@ -19,6 +19,8 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
 # IN THE SOFTWARE.
 
+import math
+
 import boto
 from boto.connection import AWSQueryConnection
 from boto.sqs.regioninfo import SQSRegionInfo
@@ -92,7 +94,7 @@ class SQSConnection(AWSQueryConnection):
         params = {'QueueName': queue_name}
         if visibility_timeout:
             params['Attribute.1.Name'] = 'VisibilityTimeout'
-            params['Attribute.1.Value'] = int(visibility_timeout)
+            params['Attribute.1.Value'] = int(math.ceil(visibility_timeout))
         return self.get_object('CreateQueue', params, Queue)
 
     def delete_queue(self, queue, force_deletion=False):
@@ -304,7 +306,7 @@ class SQSConnection(AWSQueryConnection):
         """
         params = {'MessageBody' : message_content}
         if delay_seconds:
-            params['DelaySeconds'] = int(delay_seconds)
+            params['DelaySeconds'] = int(math.ceil(delay_seconds))
 
         if message_attributes is not None:
             keys = sorted(message_attributes.keys())


thieman:
@danielgtaylor or someone: could this be reviewed, please?

signalpillar:
@thieman I think that existing solution using `int` is good enough as documentation clearly says what type is required or we can treat it as - value will be cast to `int`. 

In your case we have `3.99` as a value in my case it is a `3.1` and I don't expect it to be **4** =) 
I would agree with you that `TypeError` is more appropriate, but not sure such solution should be applied in this case.

thieman:
I disagree strongly that documentation is a sufficient type checker. If it were, I wouldn't have hit this problem and I wouldn't be here trying to fix it now. :smile: 

The point of my PR is that the current functionality does not do what the user expects it to do: if I pass `3.1` seconds to as a delay parameter and the item is made available after only `3` seconds, that is not at all what I expected and there are certain system designs (in my case, a debouncer) that fail catastrophically and lose data when this happens. Alternatively, having the item be available after `4` seconds should be fine because SQS doesn't guarantee high precision on the delay window, anyway. So even if you ask SQS to delay an item for `3` seconds, it's not guaranteed to be made available *exactly* `3` seconds later. But they *do* guarantee it will not be made available *before* `3` seconds have elapsed.

I think the ship has sailed on throwing a `TypeError` here as it would likely break a lot of existing installs. However, if we are going to be changing the user's value anyway (in your example, if you're passing `3.1`, why is it okay to be `3` but not `4`?) we should prefer to change it to something that honors the user's intent.

signalpillar:
@thieman I see your point, but the problem this code never knows the user's intent, right ? As our examples show =) 

Do you consider to validate on client side of the library ? There we know more about user expectations and more about input values range.

thieman:
I think it would be fine to delegate this responsibility to the client **if** Boto raised a `TypeError` instead of doing a cast. Once Boto starts changing the user's input without telling the user about it, it's taking the responsibility on to itself to do what the user *means* instead of what they entered. You can't blindly accept the user's malformed input and also expect them to read the docs and only send you valid input.

I'm unaware of any use case for delay queues where it would be preferable to delay for less time than the user meant rather than more time. The entire *point* of giving the delay argument is that you really do not want the item to be available until that amount of time has passed. The current functionality is inherently broken.

I, personally, am completely fine with throwing a `TypeError` if you think that's a better approach, but the current behavior is the worst of all the possible solutions.

thieman:
@signalpillar Are there any other contributors we can bring into this conversation?

To provide another example showing that mathematical rounding does not do what the user wants, consider a delay parameter of `0.49` seconds. If we're not throwing an error, then we must choose between actually delaying for `0` or `1` seconds. The user clearly does not want the item to be made available immediately, as she went through the effort of passing a delay parameter in the first place. Therefore our choice is easy, we should delay the item for `1` second.

signalpillar:
@thieman I don't know anybody.

kyleknap:
Yeah I am not sure if we can take this PR mainly because it is a change of behavior from what people may rely on currently (even though it may be unintentional). Also throwing a type error would be a breaking behavior for some people if they happen to be passing a float. I agree that throwing an error would be the best option however that would entail breaking user's runtimes. Thoughts?

thieman:
I would argue this is a bug fix. Surely changing behavior is preferred if the existing behavior is broken? The `ceil` offers the minimum disruption to end users. If you're passing an int already, nothing changes. If you're passing a float, this gives you the shortest delay possible through the SQS API that honors the intention of the delay parameter, namely "Do not make this available until this much time has passed."

I guess I'll rest my case at this point as I've been reduced to repeating myself. This is a bug. The fix is simple. Please merge the fix.

:heart: - Travis

PR #3095, opened 2015-04-13
Fixing release number in the title

joonathan:
Release number was given as 2.32.2, but it should be 2.35.2. This impacted the index page release numbering.


kyleknap:
Thanks. Merging.

PR #3101, opened 2015-04-21
[Doc] Wrong order of loading config file

dhoeric:
According to https://github.com/boto/boto/blob/a60ee4c/boto/provider.py#L277-302


kyleknap:
Thanks! Merging.

PR #3105, opened 2015-04-23
sub-second timeouts should be possible

sglebs:
timeouts can be float throughout the Python library. In boto the same should be true. Assume it is a float and not an int. So, half a second is now possible: 0.5


kyleknap:
Can you provide your use case for needing floats?  Also I think there are some other places that this needs to get updated if we accept the PR.

sglebs:
The use case is that I want an S3 timeout of 250ms. That's 1/4 of a second. You need a float for that, since the unit is in seconds and not milliseconds.

Issue #3103, opened 2015-04-22
AWS removing support for SSLv3


dannosaur:
An email received from AWS this morning indicated that they will be removing support for SSLv3 connections to S3, and likely most of their other services either at the same time, or some time in the future. Upon inspecting of the boto code, it appears that the boto.https_connection does not specify an SSL version when calling Python's ssl.wrap_socket.

I imagine a solution would simply be to add __ssl_version="TLSv1"__ to wrap_socket's arguments, but was also wondering if there was perhaps a reason even the latest/develop version does not specify this. If there isn't, I can submit a pull request with this change in?

kofic:
Thanks for opening this issue @dannosaur - we are also very dependent on boto and would like to get an answer to @dannosaur's question.

seamr:
AWS will discontinue support of SSLv3 for securing connections to S3 buckets on 12:00 AM PDT April 30, 2015 - that's in 7 days.

hassek:
This looks like a critical issue. +1 for a hotfix

dtran320:
+1 for this. Has anyone received a list of user agents requesting S3 files via SSLv3 from AWS? Waiting to hear back-- just want to see if Boto is on that list.

dtran320:
Hmm, so it seems like the SSLv23 will use TLS if you have OpenSSL 1.0.1+ installed. Can anyone verify that? 

https://docs.python.org/2/library/ssl.html

marlev:
+1
Thanks for opening this issue @dannosaur

hassek:
I just tested manually the @hzy commit connection and uploading/downlaoding a file and it works fine.

Although, as pointed out [here](https://docs.python.org/2/library/ssl.html#socket-creation) by @dannosaur if the client uses SSLv23 and has installed openssl v1.0.1+ it will be able to connect to a server using TLSv1+, if this is the case I think it would be better to instead add an exception to force the library users to update to openssl v1.0.1+ if they haven't by checking the `ssl.OPENSSL_VERSION`

dtran320:
+1 for @hassek's solution-- would be awesome if both threw some kind of DeprecationWarning or flat out raised an exception if OpenSSL is < 1.0.1

jamesls:
Hi everyone, I’d just like to give an update on boto and the impact of SSLv3 removal.

First, here’s an overview of the current state of things:

[...]

PR #3114, opened 2015-04-28
Set SSL version to TLSv1

hzy:
S3 is deprecating SSLv3, see discussion here https://github.com/boto/boto/issues/3103

### DIFF
--- a/boto/connection.py
+++ b/boto/connection.py
@@ -832,7 +832,8 @@ class AWSAuthConnection(object):
             sslSock = ssl.wrap_socket(sock, keyfile=key_file,
                                       certfile=cert_file,
                                       cert_reqs=ssl.CERT_REQUIRED,
-                                      ca_certs=self.ca_certificates_file)
+                                      ca_certs=self.ca_certificates_file,
+                                      ssl_version=ssl.PROTOCOL_TLSv1)
             cert = sslSock.getpeercert()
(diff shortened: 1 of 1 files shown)


ssynchron:
I believe this doesn't completely address the issue. Please take a look at https://github.com/ssynchron/boto/commit/b7b52ba77ed74ce6d927f4f4c074cf25a58099c7

jamesls:
Hi thanks for the pull request.  As was mentioned in #31303, ssl.PROTOCOL_SSLv23 will pick TLS provided it is supported server side.

I also believe @ssynchron is correct, boto needs to modified in at least those three places.

However, if I'm reading the [table here](https://docs.python.org/2/library/ssl.html#socket-creation) correctly then this will request TLS1 only and NOT newer versions TLS1.1 and TLS1.2 if they're available.  Is that correct?  I think ideally we should be using the highest supported version of TLS.

If we were going to go down this route, it seems like the best solution would be to still keep PROTOCOL_SSLv23, but disable SSLV3 via [ssl.OP_NO_SSLv3](https://docs.python.org/2/library/ssl.html#ssl.OP_NO_SSLv3).  The one complication is that we still need to support python2.6, and OP_NO_SSLv3 was added in python2.7.9.

hzy:
Thanks for the clarification, looks like this change isn't necessary.

PR #3117, opened 2015-04-29
Fixup section headers so underline is same length as header text.  While...

freimer:
... the standard says the underline needs to be "at least as long as the title text" some editors flag it as an error if it is not the same length.


kyleknap:
Thanks! Merging.

PR #3118, opened 2015-04-30
Remove DisableRollback from being part of the default parameter set for create_stack()

alsmith:
Remove DisableRollback from being specified, as if you want to call create_stack() like this:

    stack = aws_conn.create_stack(stack_name, template, on_failure='DELETE')

then AWS complains that DisableRollback AND OnFailure are both specified and it doesn't want to actually create the stack.

### DIFF
--- a/boto/cloudformation/connection.py
+++ b/boto/cloudformation/connection.py
@@ -224,8 +224,7 @@ class CloudFormationConnection(AWSQueryConnection):
         :rtype: dict
         :return: JSON parameters represented as a Python dict.
         """
-        params = {'ContentType': "JSON", 'StackName': stack_name,
-                'DisableRollback': self.encode_bool(disable_rollback)}
+        params = {'ContentType': "JSON", 'StackName': stack_name}
         if template_body:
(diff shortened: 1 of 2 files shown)


kyleknap:
Thanks for the PR. However, I do not think we will be able to accept this pull request because it is a change of behavior by removing it as the default. It looks like you must specify ``disable_rollback`` to avoid the error you were getting. Closing PR. If you have any other ideas please reopen.

alsmith:
The problem here is that boto always wants to specify disable_rollback: it's an implicit parameter so you can't call create_stack() without create_stack() wanting to also put a value in for disable_rollback.

That means that if you want to specify on_failure=<whatever> then AWS complains that you can't specify on_failure AND disable_rollback together and refuses to create the stack.

IMHO it is not correct to have create_stack() always want to pass a disable_rollback parameter, if the user didn't ever want to have one specified - it is indeed a change of behaviour, but one for the better.

Another way to solve this would be to see if the user specified on_failure=something and if so to then discard any disable_rollback request, but that just seems messy and non-intuitive, doesn't it?

PR #3121, opened 2015-04-30
Fix broken example in docs

collinstocks:
Must specify the primary hash key in local indexes, or you get `boto.dynamodb2.exceptions.ValidationException: ValidationException: 400 Bad Request`.


pauloubuntu:
I haven't thought about it yet. Good catch :)

collinstocks:
This is actually an issue that was fixed a while ago in the documentation in a couple other places, but this example was missed.

kyleknap:
Thanks! Merging.

PR #3122, opened 2015-04-30
``SQSConnection.get_queue`` returns None in all error-cases.

signalpillar:
Problem
-------

``SQSConnection.get_queue`` returns None in all error cases - just swallows all SQSError.

It is not possible to distinguish when missing queue with the given name from auth error.

Solution
--------

According to the [doc](http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_GetQueueUrl.html) check case when status code is 400 and error code is about missing queue.

Changes
-------

- fix itself
- small PEP8 alignments

Testing
-------
- added two additional cases 
    + error propagated in case of error different from missing queue by specified name
    + None returned only in case of error ``AWS.SimpleQueueService.NonExistentQueue``


boto/sqs/connection.py
```diff
@@ -447,21 +447,26 @@ def get_all_queues(self, prefix=''):
...
+            params['QueueOwnerAWSAccountId'] = owner_acct_id
+
+        result = None
         try:
-            return self.get_object('GetQueueUrl', params, Queue)
-        except SQSError:
-            return None
+            result = self.get_object('GetQueueUrl', params, Queue)
+        except SQSError as error:
+            if not (error.code == 'AWS.SimpleQueueService.NonExistentQueue'
```
pauloubuntu: @signalpillar I was just wondering ... NonExistentQueue is the only error that currently returns a 404 status code ? If so, I think it'd be prudent to use only the status code as long as it doesn't have another scenarios which it might happen.

signalpillar: @pauloubuntu as I understood `400` is returned also in other cases. I referred this [document](http://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/CommonErrors.html). There are multiple cases when `400` is returned too. 

So I decided to be more explicit about the case when None should be returned.

boto/sqs/connection.py
```diff
@@ -447,21 +447,26 @@ def get_all_queues(self, prefix=''):
...
+                    and error.status == 400):
+                raise
```
pauloubuntu: Instead of just raising an issue, I think you should also log it to boto's default logging strategy. Here is an example of how to do it. https://github.com/boto/boto/blob/develop/boto/dynamodb2/table.py#L454-L456

boto/sqs/connection.py
pauloubuntu: Got it :) Thanks for you quick reply

pauloubuntu:
@signalpillar I've added just a few minor comments wondering why things took that way but they are non blocking if @danielgtaylor, @kyleknap decide to merge it.

pauloubuntu:
Good job ! :+1:

boto/sqs/connection.py
signalpillar: @pauloubuntu I was thinking about the same but two things prevented from additional logging 
- `get_object` does this already 
- None of other methods in ``SQSConnection`` log as primitive methods (`get_object`, `get_status`, `get_list`) we use does this for us.

signalpillar:
@pauloubuntu regarding logging, I am just not sure =) I decided to follow current model of logging. 

Thank you for the review.

pauloubuntu:
Hummm, I see. Perhaps it's a good time to implement it. As soon as this PR gets merged I'm intended to add the boto's default logging to it.  =)  Once again thanks for your PR.

kyleknap:
Can you rebase the commits off the HEAD of develop? In the meantime, we need to verify that there is no backwards incompatible changes.

signalpillar:
@kyleknap done, though I merged develop into my branch

PR #3124, opened 2015-04-30
Build policy public

### LINKED ISSUE #3123: Make _build_policy in iam.connection public
For writing idempotent iam role policies, it's necessary to fall back to a default policy when a policy isn't defined. Right now, when calling update_assume_role_policy, it's necessary to pass in a role policy. However, in our situation, if a policy isn't defined, we want to use the default, so we need to know the default to compare against.

There's no stable way to currently get this out of boto.

ryan-lane:
Fixes #3123


boto/iam/connection.py
```diff
@@ -1102,7 +1102,7 @@ def create_instance_profile(self, instance_profile_name, path=None):
             params['Path'] = path
         return self.get_response('CreateInstanceProfile', params)
 
-    def _build_policy(self, assume_role_policy_document=None):
+    def build_policy(self, assume_role_policy_document=None):
```
signalpillar: If we publish this method lets update the docstring for it.

```
Build default policy document when `assume_role_policy_document` is not specified.

:type assume_role_policy_document: str
:return: Serialized to string JSON document of the policy
:rtype: str
```

**Not sure the docstring is complete.

ryan-lane: That docstring looks pretty accurate to me. Updated PR to add it.

signalpillar: Good. Thank you.

ryan-lane:
The default policy is also a bit wonky for this. AWS converts single item lists into strings, also the policy is missing a version string, so comparing what we put in vs what we get back is difficult. I'll update this PR to make this more reliable.

Issue #3126, opened 2015-05-01
EMR list_instances filtering by instance_group_types is broken


markreddy:
When listing instances using list_instances and providing a filter the response is returned is incorrect. This is caused by using the wrong parameter key (label) InstanceGroupTypeList instead of InstanceGroupTypes as per: http://docs.aws.amazon.com/ElasticMapReduce/latest/API/API_ListInstances.html

```python
        if instance_group_types:
            self.build_list_params(params, instance_group_types,
                                   'InstanceGroupTypeList.member')
```
Should be:
```python
        if instance_group_types:
            self.build_list_params(params, instance_group_types,
                                   'InstanceGroupTypes.member')
```

To recreate attempt to list only the MASTER node in an EMR cluster:
```python
from boto.emr.connection import EmrConnection
from boto.ec2.regioninfo import RegionInfo
conn = EmrConnection('access_key', 'secret_key', region=RegionInfo(name='region', endpoint='elasticmapreduce.eu-west-1.amazonaws.com'))
resp = conn.list_instances(cluster_id=clusterId, instance_group_types=["MASTER"])
```
resp will contain all instances in the InstanceGroup and will not be filtered to just the MASTER, as the filter key is not valid.

mtdowling:
PR merged. Closing.

PR #3127, opened 2015-05-01
Fixes Issue #3126 EMR list_instances when fitlering by instance_group_types

markreddy:
Issue #3126 fixes EMR list_instances when fitlering by instance_group_types


achiku:
+1 for this !!

kyleknap:
Could you add a test for this?

mtdowling:
I've confirmed that this is the correct format and I see that the test was updated to reflect it. Merging.

PR #3136, opened 2015-05-05
document route53 updating a record

noordzij:
It took me some time to figure out how to do a simple create or update of a route 53 dns record. So for future reference I updated the route 53 tutorial with an example...


kyleknap:
Thanks! Merging.

PR #3139, opened 2015-05-05
Fix load_from_path in boto.pyami

eshyong:
`Config#load_from_path` was leaving a file pointer dangling. Fix that by using a with statement.


eshyong:
Hi, I tried to look for test files in boto/pyami, but could not find any. If anyone would like to point me to a place where I can test this functionality, or has any suggestions on how I could get this merged, that would be great :smiley:

kyleknap:
@eshyong 
Thanks for the PR! It would also be great if you could rebase off the HEAD of develop. There are a bunch of commits that are not needed in this PR. Also it looks like the last line changed does not have the same relative indentation as before the ``self.read()`` one. Is this on purpose?

It would be great to get tests in on this as I could not find any test for it previously. You may have to start your own suite.

brutasse:
#3535 does the same thing. It would be great if one or the other could be merged to close the file descriptor more explicitly.

PR #3141, opened 2015-05-06
Fix route53 error response parsing in retry handler.

yossigo:
The Route53 response handler currently looks for retryable error codes in a `Code` HTTP header when a 400 response is received.

Looking at the documentation and running some tests against AWS shows that this is wrong, and the error code should be parsed in the response XML body.  As a result, the current retry handler basically does nothing.

### DIFF
--- a/boto/route53/connection.py
+++ b/boto/route53/connection.py
@@ -584,20 +584,25 @@ class Route53Connection(AWSAuthConnection):
         boto.log.debug("Saw HTTP status: %s" % response.status)
 
         if response.status == 400:
-            code = response.getheader('Code')
+            body = response.read()
 
-            if code:
+            # We need to parse the error first
+            err = exception.DNSServerError(
+                response.status,
+                response.reason,
+                body)
+            if err.error_code:
                 # This is a case where we need to ignore a 400 error, as
                 # Route53 returns this. See
                 # http://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DNSLimitations.html
-                if 'PriorRequestNotComplete' in code:
-                    error = 'PriorRequestNotComplete'
-                elif 'Throttling' in code:
-                    error = 'Throttling'
-                else:
+                if not err.error_code in (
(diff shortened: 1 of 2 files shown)


hbm5006:
This is awesome. I ran into the broken route53 retry logic recently and this fixed the issue for me. Would really like to see this get merged into master soon...

dkhandeshi:
Would like to see this getting merged soon - we are running into the same issue!

Aydnwylde:
This fixed my issue.  Merge would be great.

kyleknap:
@yossigo 

Can you point to the documentation in the API reference that you are referring to? I am also hesitant to completely remove the checking of the headers for ``Code`` as this has the possibility to break people if this does work for some people. Ideally I would want to keep the existing checks in place and just add  on the new checks that you have in your PR.  It would be great if you could update your PR to accommodate that.

yossigo:
@kyleknap Here is a reference to the REST Request/Response format:
http://docs.aws.amazon.com/Route53/latest/APIReference/requests-rest-responses.html

It clearly indicates how the HTTP Status Code can be used, but does not make any reference to an additional "Code" HTTP header, but to an XML body instead.

The following page describes common headers used, and makes no reference to such header. Also, note that all non-standard headers are prefixed by an "x-amz" prefix.

http://docs.aws.amazon.com/Route53/latest/APIReference/Headers-Common.html

I think it would be safe to make this change.

mdpye:
@kyleknap can I prompt any renewed interest in merging this given the unambiguous documentation links @yossigo provided?

Although I suspect it is not necessary, I would be happy to rework the PR  as you requested, if that would result in it being merged. The current behaviour is broken for the majority of people and no one want to maintain a long term form to fix it...

kyleknap:
I rereviewed it. I am convinced that we do not have to worry about the header. The documentation does not say anything and even in botocore we only look in the body: https://github.com/boto/botocore/blob/develop/botocore/parsers.py#L711. Merging.

PR #3142, opened 2015-05-06
Fix ec2containerservice

phobologic:
This should allow connection to all the endpoints for all Regions that
    ECS is in.  Also there was a bad import path for the ECS connection
    object.

    There's also an 'ecs' package, not sure which I should be using, but
    this one needed some love.


phobologic:
Ahh, I see - the ecs module is for 'ecommerce', wasn't aware of that.  Anyway, this module needed this fix.

phobologic:
I'm going to close this and re-open my fix without the merge's - not sure why those are in this pull.  I'll open a new pull soon.

PR #3143, opened 2015-05-07
Fix import path for ec2container connection

phobologic:
Also add new endpoints for ECS.

This is missing a test case, but the test case would mostly be that import works, and that seems a bit excessive.


phobologic:
Not sure if there's anything else that is required for this.  Would love to see it in boto - we have to patch manually for the time being.

klj613:
i need this

nickrowlandboxuk:
+1

jamesls:
Taking a look.

jamesls:
Pulled in your PR and added a test for this .  Thanks for the PR.

nickrowlandboxuk:
Thanks @phobologic @jamesls :+1:

klj613:
Thanks :)

PR #3147, opened 2015-05-07
Fix #2549, add tag support for ELBs

### LINKED ISSUE #2549: Boto should support ELB Tags

JaredReisinger:
Adds getting, adding, and removing tags from ELBs, both from an ELBConnection or from an individual LoadBalancer.

### DIFF
--- a/boto/ec2/elb/__init__.py
+++ b/boto/ec2/elb/__init__.py
@@ -30,6 +30,7 @@ from boto.ec2.instanceinfo import InstanceInfo
 from boto.ec2.elb.loadbalancer import LoadBalancer, LoadBalancerZones
 from boto.ec2.elb.instancestate import InstanceState
 from boto.ec2.elb.healthcheck import HealthCheck
+from boto.ec2.elb.tag import TagDescriptions
 from boto.regioninfo import RegionInfo, get_regions, load_regions
 import boto
 from boto.compat import six
@@ -756,3 +757,70 @@ class ELBConnection(AWSQueryConnection):
                                'Subnets.member.%d')
         return self.get_list('DetachLoadBalancerFromSubnets',
                              params, None)
+
+    # Tag methods
+
+    def build_tag_param_list(self, params, tags):
+        keys = sorted(tags.keys())
+        i = 1
+        for key in keys:
+            value = tags[key]
+            params['Tags.member.%d.Key' % i] = key
+            if value is not None:
+                params['Tags.member.%d.Value' % i] = value
+            i += 1
+
+    def get_all_tags(self, load_balancer_names=None):
+        """
+        Retrieve all the metadata tags associated with your ELB(s).
+
+        :type load_balancer_names: list
+        :param load_balancer_names: An optional list of load balancer names.
+
+        :rtype: list
+        :return: A list of :class:`boto.ec2.elb.tag.Tag` objects
+        """
+        params = {}
+        if load_balancer_names:
+            self.build_list_params(params, load_balancer_names,
+                                   'LoadBalancerNames.member.%d')
+
+        return self.get_object('DescribeTags', params, TagDescriptions,
+                               verb='POST')
+
+    def add_tags(self, load_balancer_names, tags):
+        """
+        Create new metadata tags for the specified resource ids.
+
+        :type load_balancer_names: list
+        :param load_balancer_names: A list of load balancer names.
+
+        :type tags: dict
+        :param tags: A dictionary containing the name/value pairs.
+                     If you want to create only a tag name, the
+                     value for that tag should be the empty string
+                     (e.g. '').
+        """
+        params = {}
+        self.build_list_params(params, load_balancer_names,
+                               'LoadBalancerNames.member.%d')
+        self.build_tag_param_list(params, tags)
+        return self.get_status('AddTags', params, verb='POST')
+
+    def remove_tags(self, load_balancer_names, tags):
+        """
+        Delete metadata tags for the specified resource ids.
+
+        :type load_balancer_names: list
+        :param load_balancer_names: A list of load balancer names.
+
+        :type tags: list
+        :param tags: A list containing just tag names for the tags to be
+                     deleted.
+        """
+        params = {}
+        self.build_list_params(params, load_balancer_names,
+                               'LoadBalancerNames.member.%d')
+        self.build_list_params(params, tags,
+                               'Tags.member.%d.Key')
+        return self.get_status('RemoveTags', params, verb='POST')
--- a/boto/ec2/elb/loadbalancer.py
+++ b/boto/ec2/elb/loadbalancer.py
@@ -25,6 +25,7 @@ from boto.ec2.elb.listener import Listener
 from boto.ec2.elb.listelement import ListElement
 from boto.ec2.elb.policies import Policies, OtherPolicy
 from boto.ec2.elb.securitygroup import SecurityGroup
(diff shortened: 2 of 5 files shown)


boto/ec2/elb/__init__.py
```diff
@@ -756,3 +757,71 @@ def detach_lb_from_subnets(self, name, subnets):
                                'Subnets.member.%d')
         return self.get_list('DetachLoadBalancerFromSubnets',
                              params, None)
+
+    # Tag methods
+
+    def _build_tag_param_list(self, params, tags):
+        keys = sorted(tags.keys())
+        i = 1
+        for key in keys:
```
signalpillar: we can use `enumerate(keys, start=1)` here

boto/ec2/elb/__init__.py
```diff
@@ -756,3 +757,71 @@ def detach_lb_from_subnets(self, name, subnets):
                                'Subnets.member.%d')
         return self.get_list('DetachLoadBalancerFromSubnets',
                              params, None)
+
+    # Tag methods
+
+    def _build_tag_param_list(self, params, tags):
```
signalpillar: We can improve a bit this method:
- make it class method or static method as it doesn't depend on the class state
- avoid side-effects - not pass `params` for the modification but return a new instance on every call
- rename to not confuse - we expect getting `list` as a result, but we get mapping. 

```
@classmethod
def _build_tag_parameters(cls, tags):
    params = {}
    for idx, key in enumerate(sorted(tags), start=1):
        params['....'] = key
        value = tags[key]
        if value is not None:
            params['...'] = value
    return params        
        
```

boto/ec2/elb/__init__.py
```diff
@@ -756,3 +757,71 @@ def detach_lb_from_subnets(self, name, subnets):
...
+        :type tags: dict
+        :param tags: A dictionary containing the name/value pairs.
+                     If you want to create only a tag name, the
+                     value for that tag should be the empty string
+                     (e.g. '').
+        """
+        params = {}
+        self.build_list_params(params, load_balancer_names,
+                               'LoadBalancerNames.member.%d')
+        self._build_tag_param_list(params, tags)
```
signalpillar: with change above we can change it to

`params.update(self._build_tag_params(tags))`

boto/ec2/elb/loadbalancer.py
```diff
@@ -417,3 +419,85 @@ def apply_security_groups(self, security_groups):
...
+
+    def get_tags(self, force=False):
+        """
+        Get the tags for this load balancer.  Adding a tag involves a round-trip
+        to the ELB service.  Tags will be cached.
+
+        :type force: bool
+        :param force: Ignore cache value and reload.
+
+        :rtype: boto.ec2.elb.tag.TagSet
```
signalpillar:
```
:class:`boto.ec2.elb.tag.TagSet`
```

tests/unit/ec2/elb/test_tag.py
```diff
@@ -0,0 +1,134 @@
...
+    connection_class = ELBConnection
+
+    def default_body(self):
+        return REMOVE_TAGS_RESPONSE
+
+    def test_remove_tag(self):
+        self.set_http_response(status_code=200)
+        lb = LoadBalancer(self.service_connection)
+        lb.name = "lb-abcd1234"
+        lb._tags = TagSet()
```
signalpillar: hm, `TagSet` which is `dict`. Name is more about type `set[Tag].
Did you consider other naming ?

signalpillar:
+1

tests/unit/ec2/elb/test_tag.py
JaredReisinger: The naming comes in part for the implementation of tagging elsewhere in boto.  Do you have a recommendation for an alternate name (like `TagDict`, say)?  I actually started with a `Tag` class, but it turns out that unlike the EC2 idea of tags, the ELB tags are name/value _only_... so a simple dict seemed the most straightforward way to go.

boto/ec2/elb/__init__.py
JaredReisinger: The name and implementation come pretty directly from `ec2/connection.py`, around line 4137.  (Which isn't to say that the EC2 implementation shouldn't change as well!)  Unlike in EC2, though, the ELB version is only ever called once... so it could simply be inlined into the calling function.

Are you imagining that this function would return a new dict containing _only_ the tag key/values, and then the caller would `update()` the actual params dict with the returned one?

boto/ec2/elb/__init__.py
JaredReisinger: Hah!... apparently, I just needed to keep reading...  ;-)

boto/ec2/elb/loadbalancer.py
JaredReisinger: Nice catch... thanks!

JaredReisinger:
Thanks for the feedback, @signalpillar!  I really do like the tag param builder change; it feels much cleaner.

tests/unit/ec2/elb/test_tag.py
signalpillar: I thought about `Tags` as some neutral option, but, again, I am not sure ;)

boto/ec2/elb/__init__.py
```diff
@@ -756,3 +757,71 @@ def detach_lb_from_subnets(self, name, subnets):
...
+    def _build_tag_parameters(tags):
+        params = {}
+        for idx, key in enumerate(sorted(tags.keys()), start=1):
+            params['Tags.member.%d.Key' % idx] = key
+            value = tags[key]
+            if value is not None:
+                params['Tags.member.%d.Value' % idx] = value
+        return params
+
+    def get_all_tags(self, load_balancer_names=None):
```
AMeng: Attempting to call `get_all_tags()` without any load balancer names results in a 400 from AWS:
```
*** BotoServerError: BotoServerError: 400 Bad Request
<ErrorResponse xmlns="http://elasticloadbalancing.amazonaws.com/doc/2012-06-01/">
  <Error>
    <Type>Sender</Type>
    <Code>ValidationError</Code>
    <Message>1 validation error detected: Value null at 'loadBalancerNames' failed to satisfy constraint: Member must not be null</Message>
  </Error>
  <RequestId>df0a9c1a-fd9b-11e4-95e7-27e321da6c35</RequestId>
</ErrorResponse>
```

As of now, the ELB [describe-tags](http://docs.aws.amazon.com/cli/latest/reference/elb/describe-tags.html) endpoint only supports getting individual ELB's tags at a time.

`load_balancer_names` should probably be a required argument instead of a keyword argument.

JaredReisinger: Will do.  (I'm currently without my laptop... but will get to this in about 1 week.  Just lucky timing I happened to catch your comment!)

JaredReisinger:
I think this is complete... ELB tags can be retrieved, set, and removed, which is the complete set of functionality for current ELB tags.  What else do I need to do to help get this feature/fix ready for boto?

kyleknap:
You should be good for now. We just need to verify the methods you added correctly reflect ELB's API via their reference docs.

PR #3149, opened 2015-05-08
Route53 TCP healthcheck

zimbatm:
Health checks with type TCP couldn't be defined because the ResourcePath was always provided


zimbatm:
Abandoning this change

PR #3150, opened 2015-05-08
Move get_location to connection.py

7heo:
Having get_location in bucket.py made no sense, since the location needs
to be known for the connection to be established.


7heo:
Merged into https://github.com/boto/boto/pull/3161

PR #3153, opened 2015-05-08
Fix #3119. Use response headers in presigned urls.

### LINKED ISSUE #3119: S3 generate signed downloading url is not using response headers
I'm trying to generate a download url for an S3 file/key and want to change the filename when the downloaded file is saved locally (using the Content-Disposition header).
```python
disposition = "attachment; filename=\"{}\"".format(file_name)
response_headers = {
    'response-content-disposition': disposition
}
key.generate_url(60 * 60, query_auth=True, response_headers=response_headers)
```

The [`key.generate_url`](https://github.com/boto/boto/blob/develop/boto/s3/key.py#L629) method uses the [`s3.connection.generate_url_sigv4` method](https://github.com/boto/boto/blob/develop/boto/s3/connection.py#L353) and this function doesn't seem to use the `response_headers` argument, so my Content-Disposition header is not used in the response.

In the normal `s3.connection.generate_url`, the response headers are [added to the query path](https://github.com/boto/boto/blob/develop/boto/s3/connection.py#L397-399) like described in the [S3 documentation](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGET.html#RESTObjectGET-requests-request-parameters).

Kind regards,
Peter

peterneyens:
Response headers are now added to the query path like in the regular
[`s3.connection.generate_url`](https://github.com/boto/boto/blob/develop/boto/s3/connection.py#L397-399).

### DIFF
--- a/boto/s3/connection.py
+++ b/boto/s3/connection.py
@@ -366,6 +366,10 @@ class S3Connection(AWSAuthConnection):
         if version_id is not None:
             params['VersionId'] = version_id
 
+        if response_headers is not None:
+            for header, value in response_headers.items():
+                params[header] = value
+
         http_request = self.build_base_http_request(method, path, auth_path,
                                                     headers=headers, host=host,
                                                     params=params)
--- a/boto/s3/connection.py
+++ b/boto/s3/connection.py
@@ -367,8 +367,7 @@ class S3Connection(AWSAuthConnection):
             params['VersionId'] = version_id
 
         if response_headers is not None:
-            for header, value in response_headers.items():
-                params[header] = value
+            params.update(response_headers)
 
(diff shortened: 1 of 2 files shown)


boto/s3/connection.py
```diff
@@ -366,6 +366,10 @@ def generate_url_sigv4(self, expires_in, method, bucket='', key='',
         if version_id is not None:
             params['VersionId'] = version_id
 
+        if response_headers is not None:
+            for header, value in response_headers.items():
```
signalpillar: we can replace `for` with `params.update(response_headers)`

cancan101:
Can we get this merged?

jamesls:
Taking a look.

jamesls:
Thanks for the pull request.

PR #3161, opened 2015-05-12
OrdinaryCallingFormat: Use location in server name

7heo:
Also moved the Location class to a different file, since it has to be
imported from both connection.py and bucket.py.


7heo:
Since it is necessary to merge all PRs (but one) into a single PR for it to pass the tests (the commits are interdependent), I closed all the PRs I opened for the dotted-bucket gsutil issue, and replace them all with the unique PR https://github.com/boto/boto/pull/3163

PR #3162, opened 2015-05-12
Fix for when AWS returns empty body for ?location

7heo:
Sometimes in my tests, AWS returns empty body for the requests with the `?location` query arguments. This patch solves the bug they trigger (the XML parser complains about the empty body not being valid XML), and falls back to default region in such cases.


7heo:
Merged into https://github.com/boto/boto/pull/3161

PR #3174, opened 2015-05-19
Add method for ListHostedZonesByName API call

phobologic:
Kind of confusing, because the API call itself is confusing.  Might be
worth having another method that ONLY returns zones with the specified
domain name, though not sure what to call it.

### DIFF
--- a/boto/route53/connection.py
+++ b/boto/route53/connection.py
@@ -136,6 +136,48 @@ class Route53Connection(AWSAuthConnection):
             e = self.get_all_hosted_zones(next_marker, zone_list)
         return e
 
+    def get_all_hosted_zones_by_name(self, dns_name=None, hosted_zone_id=None,
+                                     zone_list=None):
+        """
+        Returns a Python data structure with information about all
+        Hosted Zones defined for the AWS account.
+
+        :param string dns_name: Specifies the first dnsname to be returned in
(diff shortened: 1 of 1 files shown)


boto/route53/connection.py
```diff
@@ -136,6 +136,48 @@ def get_all_hosted_zones(self, start_marker=None, zone_list=None):
...
+        Returns a Python data structure with information about all
+        Hosted Zones defined for the AWS account.
+
+        :param string dns_name: Specifies the first dnsname to be returned in
+            the list of domains. Mostly meant to be used with hosted_zone_id.
+            If left blank, will return all dns records.
+        :param string hosted_zone_id: Should not be included in first query
+            to get_all_hosted_zones_by_name, but should be used, along with
+            dns_name when dealing with truncated lists.
+        :param list zone_list: a HostedZones list to prepend to results
```
signalpillar: we can update docstring to be more details about return type and exception raised:

```
:rtype: :class:`boto.jsonresponse.Element`
:raise :class:`boto.route53.exception.DNSServerError`: Failed to get successful result.
```

phobologic: Cool - updated, as well as updated the get_all_hosted_zones method, as that's what I based my docstring off of.

PR #3176, opened 2015-05-20
fix V4 auth in one edge case

kouk:
The headers `x-amz-copy-source` and `x-amz-copy-source-range`, which are sent in the copy part S3 operation, sort differently than the header/value strings that are included in the canonical request for the V4 signature. This is because the colon character which joins the header name and the header value sorts after the dash character in `x-amz-copy-source-range`. For example python sorts the list `['x-amz-copy-source:some/key', 'x-amz-copy-source-range:bytes=0-999']` in reverse. This leads to an incorrect signature being calculated. This PR fixes the problem by sorting the header/value strings only according to the prefix before the colon.

### DIFF
--- a/boto/auth.py
+++ b/boto/auth.py
@@ -377,7 +377,7 @@ class HmacAuthV4Handler(AuthHandler, HmacKeys):
             else:
                 c_value = ' '.join(raw_value.strip().split())
             canonical.append('%s:%s' % (c_name, c_value))
-        return '\n'.join(sorted(canonical))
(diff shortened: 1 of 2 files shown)


tipabu:
FWIW, this affects more than just `x-amz-copy-source` and `x-amz-copy-source-range`; [arbitrary metadata are affected as well](https://gist.github.com/tipabu/3fb21885932c5d795fb8#file-boto-error-L42-L43). Attempting to do the same thing on this branch will successfully create the key.

PR #3181, opened 2015-05-21
Remove S3HmacAuthV4Hander::host_header() function

jasuarez:
Use instead parent version, HmacAuthV4Handler::host_header().

Fixes [boto/boto#2623].


bacongobbler:
I can reproduce this with the following script using [minio](https://github.com/minio/minio) as the test server:

```
#!/usr/bin/env python

import os
import sys

import boto
from boto.s3.connection import OrdinaryCallingFormat

os.environ["S3_USE_SIGV4"] = "true"

name = "foo"

conn = boto.connect_s3(
    host="s3-us-east-1.192.168.0.10"),
    port=9000),
    is_secure=False,
    calling_format=OrdinaryCallingFormat())

if conn.lookup(name) is None:
	conn.create_bucket(name)

```

harshavardhana:
:+1:

PR #3183, opened 2015-05-21
updates the default ttl to be more consistent with the default for the Record class

Rdbaker:
I noticed that the default ttl for the Record class didn't align with the default ttl for creating or changing a record in zone.py. This update aligns zone.py to have the same default ttl as the Record class does.

### DIFF
--- a/boto/route53/zone.py
+++ b/boto/route53/zone.py
@@ -22,7 +22,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
 # IN THE SOFTWARE.
 
-default_ttl = 60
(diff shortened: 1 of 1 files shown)


mtdowling:
I think changing the TTL so drastically would be an unexpected change to users who have been relying on the current default of 60. I do like that you updated the method to use the `default_ttl` though. Can you update this PR to not change the default TTL but to use the `default_ttl` variable as the default?

Rdbaker:
The reason it was changed to 600 is to be consistent with the web console. If we want a strict difference between the web console and the API I'm more than happy to keep it as 60, but that's just something that I thought could be consolidated between the two interfaces for aws

PR #3184, opened 2015-05-23
Retain literal wildcard characters in CF invalidation paths

wnagele:
This is to support recently introduced wildcards for partial path matching in CF (see: http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Invalidation.html).
Without this any wildcard path provided is encoded and the wildcards do not work when encoded.

### DIFF
--- a/boto/cloudfront/invalidation.py
+++ b/boto/cloudfront/invalidation.py
@@ -68,10 +68,10 @@ class InvalidationBatch(object):
         self.paths[k] = v
 
     def escape(self, p):
-        """Escape a path, make sure it begins with a slash and contains no invalid characters"""
+        """Escape a path, make sure it begins with a slash and contains no invalid characters. Retain literal wildcard characters."""
         if not p[0] == "/":
             p = "/%s" % p
-        return urllib.parse.quote(p)
+        return urllib.parse.quote(p, safe = "/*")
 
     def to_xml(self):
         """Get this batch as XML"""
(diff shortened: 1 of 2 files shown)


wnagele:
How to get this merged?

kyleknap:
Thanks for the PR. Can you add a test for this update?

wnagele:
wrong label - Route53? this is for CloudFront. i'll add a test.

wnagele:
Done and rebased against current develop HEAD.

boto/cloudfront/invalidation.py
```diff
@@ -68,10 +68,11 @@ def __setitem__(self, k, v):
         self.paths[k] = v
 
     def escape(self, p):
-        """Escape a path, make sure it begins with a slash and contains no invalid characters"""
+        """Escape a path, make sure it begins with a slash and contains no invalid characters. Retain literal wildcard characters."""
         if not p[0] == "/":
             p = "/%s" % p
-        return urllib.parse.quote(p)
+        p_quoted = urllib.parse.quote(p)
+        return p_quoted.replace('%2A', '*')
```
kyleknap: I would prefer to make the wildcard a safe character via the quote method: https://docs.python.org/2/library/urllib.html#urllib.quote

A little safer than replacing all ``%2A``.

kyleknap:
Thanks for rebasing! I had one more comment that would be great if you could address and it will be good to merge.

wnagele:
Sure - thought about using that too - wasn't sure which is prettier. Matter of preference I guess :). Either way - changed and rebased.

kyleknap:
Awesome. Thanks! Merging.

PR #3186, opened 2015-05-25
Make EMR endpoints consistent with AWS API doc

achiku:
EMR endpoints are not consistent with AWS documentation. I'm not sure why only EMR is in this style, but this patch makes a bit less confusing when looking at endpoints.

- http://docs.aws.amazon.com/general/latest/gr/rande.html#emr_region

### DIFF
--- a/boto/endpoints.json
+++ b/boto/endpoints.json
@@ -194,16 +194,16 @@
         "eu-central-1": "elasticloadbalancing.eu-central-1.amazonaws.com"
     },
     "elasticmapreduce": {
-        "ap-northeast-1": "ap-northeast-1.elasticmapreduce.amazonaws.com",
-        "ap-southeast-1": "ap-southeast-1.elasticmapreduce.amazonaws.com",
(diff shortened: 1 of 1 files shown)


achiku:
ping @jamesls @mtdowling

jamesls:
The reason for this is that there's a bug in python2.6 where we can't parse the subjectAltName of the cert so we have to use whatever's in the common name:

```
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number:
            45:ef:9b:3b:01:7c:36:9d:a5:f5:52:b8:07:19:f7:28
        Signature Algorithm: sha1WithRSAEncryption
        Issuer: C=US, O=VeriSign, Inc., OU=VeriSign Trust Network, OU=Terms of use at https://www.verisign.com/rpa (c)10, CN=VeriSign Class 3 Secure Server CA - G3
        Validity
```

So we won't be able to take this change, otherwise customers would be broken if they're using python2.6.  Thanks though for the pull request.

achiku:
@jamesls Oh, I see it. Thanks!

PR #3187, opened 2015-05-25
Allow passing recource_records to rrset add_change

lbosque:
Creating a new Record from scratch allows passing a resource_records
array with the possible values of the record set.
However invoking
add_change method over the ResourceRecordSets doesn't.


lbosque:
Moved to a different branch than develop:
https://github.com/boto/boto/pull/3191

PR #3200, opened 2015-05-30
ResultProcessor.get_results_from_bucket not incrementing self.num_files

mennis:
and miscellaneous clean up:
Mutable default arguments
Declaring variables and then immediately assigning to them is usually just extra work
Unused variables removed
Semicolons scattered about
Documentation Typos
Duplicate Declaration


kyleknap:
Can you break these up these changes into separate PRs for each change? There are a lot of changes to track here.

PR #3201, opened 2015-05-31
subclassing HTTPS rather than HTTP object. simply specifying port 443…

jtnimoy:
… is not enough to get an HTTPSConnection. now works on GAE. Please consider subclassing HTTPSConnection in more parts of this file.

This answers a handful of forum posts where people were not able to talk to AWS from Google App Engine, across various versions. I assume there were more errors in other domains, and don't have the knowledge or resources to unit-test the rest of the object in a state that subclasses HTTPSConnection. I hope someone out there does. I was using python 2.7.

### DIFF
--- a/boto/https_connection.py
+++ b/boto/https_connection.py
@@ -83,7 +83,7 @@ def ValidateCertificateHostname(cert, hostname):
     return False
 
 
-class CertValidatingHTTPSConnection(http_client.HTTPConnection):
+class CertValidatingHTTPSConnection(http_client.HTTPSConnection):
     """An HTTPConnection that connects over SSL and validates certificates."""
 
     default_port = http_client.HTTPS_PORT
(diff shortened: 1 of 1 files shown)


kyleknap:
Can you link to some of the forum posts that you were mentioning that run into this issue? Any more information you could provide would be great so that we can get a better understanding.

spukst3r:
We've run into this issue and I can confirm that this PR will fix the GAE's httplib version behaviour with boto

spukst3r:
The issue happens because Google's libraries figure out which protocol to use by class variable, not by port: [source](https://code.google.com/p/googleappengine/source/browse/trunk/python/google/appengine/dist27/gae_override/httplib.py?r=400#340), so it's essential to derive from correct class to force using HTTPS protocol

hayorov:
Please, accept this PR.

hayorov:
Up!

WeilongWu-Reflektion:
Hi,
I am running into the same issue with the tutorial code  https://blog.mturk.com/tutorial-getting-started-with-mturk-and-python-using-boto-452fb0243a30#.244n3xwmb

on Mac 10.12
python version: 2.7
boto: 2.46.1

any help would be appreciated !

PR #3202, opened 2015-06-01
Documentation fix for kms create grant

gene1wood:
the `create_grant` call doesn't actually allow the use of key aliases. If you use a key alias, AWS returns an exception

Here's code to reproduce this error. To test it you'll need to set the `kms_key` value to your KMS key's alias and set the `grantee_principle` to the ARN of an IAM role that exists

```
import boto.kms
kms = boto.kms.connect_to_region('us-east-1')
kms_key="alias/mykeyalias"
grantee_principal = 'arn:aws:iam::123456789012:role/ExampleRole'
operations = ['Decrypt']
context = {'EncryptionContextSubset': {'foo': 'bar'}}
result = kms.create_grant(key_id = kms_key, 
                          grantee_principal = grantee_principal,
                          operations=operations, 
                          constraints=context)
```

Expected results would be to have the grant created.

Actual results are the exception

```
boto.kms.exceptions.InvalidArnException: InvalidArnException: 400 Bad Request
{u'message': u'Key Aliases are not supported for this operation.', u'__type': u'InvalidArnException'}
```


jamesls:
Would you mind rebasing against develop.  The PR has several merge commits in the PR.  Ideally there's just a single commit fixing the doc.

gene1wood:
@jamesls Sure thing, I've closed this PR and opened a new one with this change in #3883

PR #3204, opened 2015-06-01
Cursor support in cloudsearch2

azreda:
In order to perform deep paging on cloudsearch, the API provides a cursor based approach.
http://docs.aws.amazon.com/cloudsearch/latest/developerguide/paginating-results.html

This is a patch to cloudsearch2 to support this cursor based approach.


kyleknap:
Thanks for the PR. Can you please add a set of tests for the feature you are adding?

kyleknap:
Also can you rebase off of the head of develop and repush your branch?

natecode:
While this patch looks overall good, it has a bug. I ran it on an index and it got in an infinite loop during paging. I think it should raise a StopIteration at some point.

The one in this other PR does work for the same case (after adding my change listed there):

https://github.com/boto/boto/pull/2815

PR #3207, opened 2015-06-02
Update cloudwatch logs endpoints

kyleknap:
cc @jamesls @mtdowling


kyleknap:
Update based on the public docs: http://docs.aws.amazon.com/general/latest/gr/rande.html#cw_region

jamesls:
:shipit:

PR #3214, opened 2015-06-05
Adds __enter__ and __exit__ methods to AWSAuthConnection

chrishenzie:
This allows users to use connections like this:
```python
import boto.ec2

with boto.ec2.connect_to_region('us-west-2',
                                aws_access_key_id=<AWS_ACCESS_KEY_ID>,
                                aws_secret_access_key=<AWS_SECRET_ACCESS_KEY>) as conn:
    conn.do_some_stuff()
```
and the connection will be automatically closed for them.

### DIFF
--- a/boto/connection.py
+++ b/boto/connection.py
@@ -574,6 +574,12 @@ class AWSAuthConnection(object):
     def __repr__(self):
         return '%s:%s' % (self.__class__.__name__, self.host)
 
+    def __enter__(self):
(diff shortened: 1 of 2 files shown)


mtdowling:
Thanks for the PR. We encourage developers to reuse clients when sending multiple requests. I think adding support for with would encourage more ad hoc creation of clients than necessary.

PR #3228, opened 2015-06-12
Fix misnamed Stack.stack_name_reason attribute

### LINKED ISSUE #3227: Misnamed attribute in `Stack`: `stack_name_reason` and `stack_status_reason`
I think these two attributes have been mixed up:

- https://github.com/boto/boto/blob/d14d5ac260a811aff3e1e1f1b1801b1bae1920bb/boto/cloudformation/stack.py#L20
- https://github.com/boto/boto/blob/d14d5ac260a811aff3e1e1f1b1801b1bae1920bb/boto/cloudformation/stack.py#L61

It doesn't make sense for the stack name to have a "reason"; it does make sense for the stack status, though. Also, if there is no status reason, `self.stack_status_reason` never gets defined, which means that code which needs to do some handling of `Stack` instances needs to be very careful, or else one will get errors like `'Stack' object has no attribute 'stack_status_reason'` (whereas it should just be `None`).

I think the solution is to replace `self.stack_name_reason` with `self.stack_status_reason` in the constructor. I'll throw up a pull request with these changes.

larsbutler:
Rename this attribute to `stack_status_reason`.

Fixes #3227.

### DIFF
--- a/boto/cloudformation/stack.py
+++ b/boto/cloudformation/stack.py
@@ -16,8 +16,8 @@ class Stack(object):
         self.tags = []
         self.stack_id = None
         self.stack_status = None
+        self.stack_status_reason = None
         self.stack_name = None
-        self.stack_name_reason = None
         self.timeout_in_minutes = None
 
     def startElement(self, name, attrs, connection):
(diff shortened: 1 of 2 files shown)


larsbutler:
Superseded by https://github.com/boto/boto/pull/3515. Closing. =)

Issue #3232, opened 2015-06-15
cannot connect to ipv6 addresses


yehudasa:
Doing the following:

```python
host = <ipv6 address>

boto.connect_s3(aws_access_key_id = access_key,
                aws_secret_access_key = secret_key,
                host=host,
                port=port,
                is_secure=False,               # uncomment if you are not using ssl
                calling_format = boto.s3.connection.OrdinaryCallingFormat(),
                )
```
results in:

Fails to connect with: [Errno -2] Name or service not known

Note that when specifying a hostname that maps into an ipv6 address it connects successfully.

Issue seems to be at boto/connection.py, where it unconditionally does this:

        host = host.split(':', 1)[0]

PR #3242, opened 2015-06-23
support IPV6 addresses when creating connections

alfredodeza:
Fixes issue #3232 by properly detecting the different variations of IPV6 addresses (with ports and brackets, just brackets, and no brackets).

The thorough test cases ensure that backwards compatibility is kept (IPV4 hosts will work just as before).

Since there is a chance the detection code will be needed for other parts of the code, they were split out into `boto/utils.py`

For some reason I couldn't get pypy tests started, but everything else passed with tox (2.6, 2.7, 3.3, and 3.4)

### DIFF
--- a/boto/utils.py
+++ b/boto/utils.py
@@ -1056,7 +1056,7 @@ def host_is_ipv6(hostname):
     Detect (naively) if the hostname is an IPV6 host.
     Return a boolean.
     """
-    # empty strins or anything that is not a string is automatically not an
+    # empty strings or anything that is not a string is automatically not an
     # IPV6 address
     if not hostname or not isinstance(hostname, str):
         return False
--- a/boto/utils.py
+++ b/boto/utils.py
@@ -1049,3 +1049,41 @@ class RequestHook(object):
     """
     def handle_request_data(self, request, response, error=False):
         pass
+
+
+def host_is_ipv6(hostname):
+    """
+    Detect (naively) if the hostname is an IPV6 host.
+    Return a boolean.
+    """
+    # empty strins or anything that is not a string is automatically not an
+    # IPV6 address
+    if not hostname or not isinstance(hostname, str):
+        return False
+
+    if hostname.startswith('['):
+        return True
+
+    if len(hostname.split(':')) > 2:
+        return True
+
+    # Anything else that doesn't start with brackets or doesn't have more than
+    # one ':' should not be an IPV6 address. This is very naive but the rest of
+    # the connection chain should error accordingly for typos or ill formed
+    # addresses
+    return False
+
+
+def parse_host(hostname):
+    """
+    Given a hostname that may have a port name, ensure that the port is trimmed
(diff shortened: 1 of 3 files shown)


mtdowling:
Is this for Amazon S3 or an S3 type emulated service?

alfredodeza:
@mtdowling good question, our specific use case is for an emulated service. However, since boto had some initial work for IPV6 support (as seen here in the release notes https://boto.readthedocs.org/en/latest/releasenotes/v2.32.0.html#boto-v2-32-0 ) I believed that further work to support IPV6 connections, regardless of the source of the service would be OK.

boto/utils.py
```diff
@@ -1049,3 +1050,39 @@ class RequestHook(object):
...
+    # IPV6 address
+    if not hostname or not isinstance(hostname, str):
+        return False
+
+    if hostname.startswith('['):  # assume we need to split on possible port
+        hostname = hostname.split(']:', 1)[0]
+    # strip leading/trailing brackets so inet_pton understands the hostname
+    hostname = hostname.strip('[]')
+    try:
+        socket.inet_pton(socket.AF_INET6, hostname)
```
mtdowling: I'm not sure we should use this here-- support for this on Windows was only added in Python 3.4: https://docs.python.org/3.4/library/socket.html#socket.inet_pton

mtdowling:
I wonder if we need to even use the socket.inet_pton function here. I'd prefer not to use it since support for this function was only added in Python 3.4. Maybe I'm oversimplifying it, but it seems like we could determine if it is IPv6 with the following checks:

1. Starts with "[" and has a closing "]" -> yes
2. Contains more than one ":" -> yes

alfredodeza:
@mtdowling the problem with going that route is that it would be less robust, but I think in this specific case we could just treat input as "this looks like IPV6" and not performing full validation on it. I will make a few changes to remove `inet_pton` and keep the test cases.

alfredodeza:
As discussed I removed `inet_pton` and use a more naive approach, all tests pass, just like before.

boto/utils.py
```diff
@@ -47,6 +47,7 @@
 import tempfile
 import random
 import smtplib
+import socket
```
mtdowling: I don't think this is needed anymore

boto/utils.py
```diff
@@ -1049,3 +1050,41 @@ class RequestHook(object):
     """
     def handle_request_data(self, request, response, error=False):
         pass
+
+
+def host_is_ipv6(hostname):
+    """
+    Detect if the hostname is an IPV6 host using the socket module. Return
```
mtdowling: This docblock is out of date now

mtdowling:
Thanks for updating the PR. There are a couple of tiny changes that I suggested. After those are done, I'll merge.

mtdowling:
Thanks!

alfredodeza:
thank you for guiding me into the best fix! 

<3

PR #3248, opened 2015-06-29
use boto.config.getint() for max_retry_delay

jhriggs:
The `max_retry_delay` config setting is used within calls to `min()`, e.g.:

```time.sleep(min(2 ** i, boto.config.get('Boto', 'max_retry_delay', 60)))```

When `max_retry_delay` is set in the config file, the *string* value returned by `boto.config.get()` will never be less than the other integer value supplied (`2 ** i` in this case) and will never be used. So, it is impossible to set the `max_retry_delay` value via config file right now.

This replaces `boto.config.get()` with `boto.config.getint()`.

### DIFF
--- a/boto/connection.py
+++ b/boto/connection.py
@@ -920,7 +920,7 @@ class AWSAuthConnection(object):
         while i <= num_retries:
             # Use binary exponential backoff to desynchronize client requests.
             next_sleep = min(random.random() * (2 ** i),
-                             boto.config.get('Boto', 'max_retry_delay', 60))
+                             config.getint('Boto', 'max_retry_delay', 60))
             try:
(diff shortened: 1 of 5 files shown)


kyleknap:
Can you add a test for the update you are making? When there are bugs we like to add test to ensure it does not happen again in the future.

jhriggs:
@kyleknap: I would be happy to, but I'm not exactly sure how we would write a test case for this. Each time a developer retrieves the `max_retry_delay` setting from the boto config — currently 6 places spread throughout the codebase — it needs to read as an int. How do/can we write test cases for future code? The same would be true for other config values like `http_socket_timeout`, `num_retries`, etc.

PR #3249, opened 2015-06-29
Fix ECS connect_to_region()

jamesls:
This pulls in https://github.com/boto/boto/pull/3143 and adds a test, and makes the implementation consistent with other services connect_to_region() functions.

cc @kyleknap @mtdowling


kyleknap:
:ship:

PR #3251, opened 2015-06-30
respect attribute from schema when setting local and global index keys

j-hao:
table.describe doesn't respect data types returned from DB when setting up local and global index key (always set to STRING) . Please see test. Without this patch, self.users.indexes[1].definition() would be
```
{ 'AttributeName': 'username', 'AttributeType': 'S'},
{ 'AttributeName': 'date_joined', 'AttributeType': 'S'}
```
instead of correct one:
```
{ 'AttributeName': 'username', 'AttributeType': 'S'},
{ 'AttributeName': 'date_joined', 'AttributeType': 'N'}
```

### DIFF
--- a/boto/dynamodb2/table.py
+++ b/boto/dynamodb2/table.py
@@ -280,7 +280,7 @@ class Table(object):
 
         return schema
 
-    def _introspect_all_indexes(self, raw_indexes, map_indexes_projection):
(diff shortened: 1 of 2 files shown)


twirrim:
Can we get an update on this?  Looks like the patch has been sitting waiting to be pulled since June and waiting for review for just shy of 3 months.

jamesls:
I'll take a look.

Issue #3089, opened 2015-04-10
How to specify Frankfurt as S3 Bucket destinaion?


joyofdata:
Hello!

As described in the [S3 section of the tutorial] I have to specify the location of a new bucket as follows:

`conn.create_bucket('mybucket', location=Location.EU)

But I would like to specifically store the bucket in Frankfurt, which is not among the available options though:

```
> from boto.s3.connection import Location
> print("\n".join([l for l in dir(Location) if l[0] != "_"]))

APNortheast
APSoutheast
APSoutheast2
CNNorth1
DEFAULT
EU
SAEast
USWest
USWest2
```

Is there a way to accomplish that or do I have to wait for a future version of boto?

Kind Regards

Raffael

charleswhchan:
Compare the [Location](https://github.com/boto/boto/blob/5e5dc4c6b31de3331c6af918b2307869988e5a73/boto/s3/connection.py#L138) object and [list of S3 endpoints](http://docs.aws.amazon.com/general/latest/gr/rande.html#s3_region), It looks like 'EU (Frankfurt)' is equivalent to 'eu-central-1'.

Likely needs to be added as it is does not yet exist in the code.

grahamlyons:
Looks like this can be closed.

PR #3254, opened 2015-06-30
Add EUCentral1 (Frankfurt) to S3 Location

HarryRybacki:
## Purpose
Add the EUCentral1 Region to S3 `Location`

## Changes
Updated `Location` object, S3 tutorial docs, and added a test.

Closes-issue: #3089

### DIFF
--- a/tests/integration/s3/test_connect_to_region.py
+++ b/tests/integration/s3/test_connect_to_region.py
@@ -57,6 +57,11 @@ class S3SpecifyHost(unittest.TestCase):
         self.assertEquals('s3.amazonaws.com', connection.host)
         self.assertIsInstance(connection, S3Connection)
 
+    def testSuccessWithDefaultEUCentral1(self):
+        connection = connect_to_region('eu-central-1')
+        self.assertEquals('s3.eu-central-1.amazonaws.com', connection.host)
+        self.assertIsInstance(connection, S3Connection)
+
     def testDefaultWithInvalidHost(self):
         connect_args = dict({'host':''})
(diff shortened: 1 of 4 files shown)


tests/integration/s3/test_connect_to_region.py
```diff
@@ -57,6 +57,11 @@ def testSuccessWithDefaultUSEast1(self):
         self.assertEquals('s3.amazonaws.com', connection.host)
         self.assertIsInstance(connection, S3Connection)
 
+    def testSuccessWithDefaultEUCentral1(self):
```
jamesls: This test doesn't exercise the new S3 location constant you've added (i.e this test passes without your changes to s3/connection.py).

It would be great to have a test that verifies we can use the new constant to create a bucket in `eu-central-1`.

HarryRybacki:
@jamesls I agree. I didn't see any location based tests for bucket creation. If I just missed them and you can point me toward them or alternatively give me a skeleton I'd be happy to write tests to verify EUCentral1 and the others.

HarryRybacki:
@jamesls -- I added a test to ensure the new location works with `s3connection.create_bucket()`

Two questions:
1. Would you like me to write similar tests for the other locations?
2. I tried to confirm the bucket's location via `bucket.get_location()` but it kept throwing nasty errors. Upon further inspection I noted that `bucket.get_location()` itself isn't tested. Is this a result of the test setup helpers boto uses or something else?

HarryRybacki:
Re-pinging @jamesls

jamesls:
Looks good, thanks for updating.

PR #3255, opened 2015-06-30
Update boto3 wording now that it's 1.0

jamesls:
cc @kyleknap @mtdowling @petermoon


mtdowling:
LGTM, though I wonder if it would be better to place this disclaimer below the introduction of what Boto actually does.

kyleknap:
Looks good. :ship:

PR #3276, opened 2015-07-21
Update queue.py

ph87:
Region `cn-north-1` is a special region located in China, the partition part of the arn is valued to 'aws-cn'.


rayluo:
We will also need a test case for this one.

Issue #3277, opened 2015-07-21
AWS GovCloud support for Cloudtrail


gburiola:
CloudTrail has been available in GovCloud since Dec 16, 2014

http://aws.amazon.com/about-aws/whats-new/2014/12/16/aws-cloudtrail-is-now-available-in-the-aws-govcloud-us-region/

http://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-endpoints.html

stevenbrichards:
Can be closed, PR has been merged

PR #3278, opened 2015-07-21
add cloudtrail support to us-gov-west-1 [issue #3277]

gburiola:
CloudTrail has been available in GovCloud since Dec 16, 2014

Fixes issue #3277

http://aws.amazon.com/about-aws/whats-new/2014/12/16/aws-cloudtrail-is-now-available-in-the-aws-govcloud-us-region/
http://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-endpoints.html


stevenbrichards:
Looks good :+1:

PR #3283, opened 2015-07-25
Add resource_records parameter to add_change, required by Route53 rec…

cliffano:
…ord deletion.

Without resource_records, committing a route53 record DELETE in a change batch results in 

```
boto.route53.exception.DNSServerError: DNSServerError: 400 Bad Request
<?xml version="1.0"?>
<ErrorResponse xmlns="https://route53.amazonaws.com/doc/2013-04-01/"><Error><Type>Sender</Type><Code>MalformedInput</Code><Message>Unexpected list element termination</Message></Error><RequestId>some-id</RequestId></ErrorResponse>
```

Which turned out to be caused by boto sending an XML body with an empty ResourceRecord element:

```
<?xml version="1.0" encoding="UTF-8"?>
<ChangeResourceRecordSetsRequest xmlns="https://route53.amazonaws.com/doc/2013-04-01/">
 <ChangeBatch>
     <Comment>None</Comment>
     <Changes>
         <Change>
                <Action>DELETE</Action>
                <ResourceRecordSet>
                    <Name>some.zone.</Name>
                    <Type>CNAME</Type>
                    <TTL>600</TTL>
                    <ResourceRecords>
                    </ResourceRecords>
                </ResourceRecordSet>
            </Change>
        </Changes>
    </ChangeBatch>
</ChangeResourceRecordSetsRequest>
```

This patch adds resource_records parameter to be added to the Record.

The record was deleted successfully when resource_records is added to the change, and obviously the value should match the value of the record to be deleted.

### DIFF
--- a/boto/route53/record.py
+++ b/boto/route53/record.py
@@ -70,7 +70,7 @@ class ResourceRecordSets(ResultSet):
     def add_change(self, action, name, type, ttl=600,
                    alias_hosted_zone_id=None, alias_dns_name=None, identifier=None,
                    weight=None, region=None, alias_evaluate_target_health=None,
-                   health_check=None, failover=None):
+                   health_check=None, failover=None, resource_records=None):
         """
         Add a change request to the set.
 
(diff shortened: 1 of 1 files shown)


rayluo:
The change looks good. Please provide a test case and we are good to go.

cliffano:
@rayluo Thanks for checking, I've added the tests, specific to the code change made in this issue.

cliffano:
@jamesls @rayluo Who is usually involved in reviewing an issue?

cliffano:
@jamesls @rayluo With boto3 out there, would you rather have this PR closed?

PR #3288, opened 2015-07-30
Fix incorrect passing of session token for swf

kyleknap:
The positional arguments did not match up so the session token was not passed to the ``AWSAuthConnection`` properly.

cc @jamesls @mtdowling @rayluo


mtdowling:
Looks like this cause some related test failures. Can you take a look?

PR #3291, opened 2015-08-03
Update ec2_tut.rst

elainexmas:
The parameter in `run_instances` representing security groups should be `security_group_ids`, not `security_groups`, per the docs for [ec2.connection](http://boto.readthedocs.org/en/latest/ref/ec2.html#module-boto.ec2.connection).


mtdowling:
You can supply `security_group_ids` or `security_groups` when calling `run_instances`. `security_groups` is used defined as "The names of the EC2 classic security groups with which to associate instances", while `security_group_ids` is defined as "The ID of the VPC security groups with which to associate instances.". Perhaps a better change would be mention that there are two parameters that each have a different purpose.

PR #3292, opened 2015-08-04
add missing 'ap-northeast-1' region for lambda

nexusz99:
Now aws lambda supports tokyo region. but boto's endpoint doesn't have tokyo region for lambda.


kyleknap:
Thanks for the pull request. Merging.

PR #3293, opened 2015-08-05
Updates cloudfront distribution documention

adregan:
time.time always returns a float (https://docs.python.org/2/library/time.html#time.time) and this argument expects an int. A float will cause a malformed policy error from cloudfront.

This pull request suggests wrapping the documentation's suggestion of using `time.time() + duration` with int() to avoid this error.


kyleknap:
Thanks for the pull request. Looks good. Merging.

PR #3295, opened 2015-08-06
Avoiding exception in clean-up file removal

rbp:
If Key.get_contents_to_filename() fails for some reason, it tries to remove `filename` as a clean-up step, before actually raising the exception. However, if _that_ call also raises a second exception (for instance, if the open() call a few lines above fails), the latter gets raised, rather than the former.

This PR silently ignores errors in the clean-up call.


kyleknap:
Can you add a test to ensure that this issue is fixed and will stay fixed in the future?

PR #3308, opened 2015-08-20
Fix 3264: Add documentation for query_auth

stevenbrichards:
Add missing documentation for `query_auth` param in S3, also add some periods to documentation to keep them the same style.


kyleknap:
Thanks. Looks good. Merging

PR #3310, opened 2015-08-20
Fix 3304: Encode bytestring to handle unicode

stevenbrichards:
See: #3304

### DIFF
--- a/boto/glacier/utils.py
+++ b/boto/glacier/utils.py
@@ -77,6 +77,8 @@ def chunk_hashes(bytestring, chunk_size=_MEGABYTE):
     for i in range(chunk_count):
         start = i * chunk_size
         end = (i + 1) * chunk_size
+        # Encode bytestring slice into 'utf-8' to avoid unicode issues see #3304
+        bytestring = bytestring.decode().encode('utf-8')
         hashes.append(hashlib.sha256(bytestring[start:end]).digest())
     if not hashes:
         return [hashlib.sha256(b'').digest()]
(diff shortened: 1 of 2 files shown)


stevenbrichards:
Fails due to Python 3 and it no longer having the `encode()` function, will modify to fix this. Fails with:
`AttributeError: 'bytes' object has no attribute 'encode'`

stevenbrichards:
Change worked (git squashed my commits too), will add some tests to verify it can handle unicode parts

stevenbrichards:
Needs PR review

jamesls:
I'd like to understand the root cause here.  If I'm understanding the original issue, this PR suggests that unicode is being introduced somewhere in this stack trace:

```
utils.chunk_hashes()
chunk_hashes()
_upload_chunk() 
_process_chunk()
_start_upload_threads()
ConcurrentUploader.upload()
concurrent_create_archive_from_file()
```


However, the fileobj is opened in binary mode so I wouldn't be expected unicode to show up anywhere: https://github.com/boto/boto/blob/develop/boto/glacier/concurrent.py#L225

How's this happening?

stevenbrichards:
I'm not sure why the unicode is appearing as I can't see the content's of the user's file they're uploading, but it happens in `utils.chunk_hashes()` with the `hashlib.sha256` function. I can try and trace my way back with the bytestream argument and see where that comes from.

stevenbrichards:
So testing locally more with a unicode file I can't recreate it. I guess we'll need to dig deeper back into the issue.

stevenbrichards:
@jamesls a place where I'm able to recreate this is by calling `compute_hashes_from_fileobj(file)` with a file containing unicode (even opened in binary mode) if this line is executed:
```
if not isinstance(chunk, bytes):
                chunk = chunk.encode(getattr(fileobj, 'encoding', '') or 'utf-8')
```
Haven't traced it all the way back to see if it's being called, but it's one of the few `encode()` calls made in the whole glacier directory that could cause this error.

PR #3320, opened 2015-08-28
Do not ignore UnprocessedKeys in dynamodb2.Table batch_get

### LINKED ISSUE #2214: UnprocessedKeys are dropped in table.py (major bug)

dcosson:
Fixes #2214 with a test

### DIFF
--- a/boto/dynamodb2/table.py
+++ b/boto/dynamodb2/table.py
@@ -1581,7 +1581,7 @@ class Table(object):
             })
             results.append(item)
 
-        raw_unproccessed = raw_results.get('UnprocessedKeys', {})
+        raw_unproccessed = raw_results.get('UnprocessedKeys', {}).get(self.table_name, {})
 
         for raw_key in raw_unproccessed.get('Keys', []):
             py_key = {}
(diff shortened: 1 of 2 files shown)


dcosson:
@danielgtaylor or other maintainers, any chance of merging this?

shiyangcao:
bumping dcosson's comment:

@danielgtaylor or other maintainers, any chance of merging this?

Issue #2852, opened 2015-01-02
s3 generate_url for multipart uploads


soulseekah:
There seems to currently be no way to generate multipart upload URLs for the whole of Mulitpart Uploads process in S3. It is impossible, in particular, to sign an initiate URL as given here http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadInitiate.html. The string to sign contains `/path?uploads` which is impossible to achieve in [`generate_url`.](https://github.com/boto/boto/blob/develop/boto/s3/connection.py#L390).

Merely appending `?uploads` the key doesn't work because of escaping, which is understandable.

The string to sign always contains `/path?uploads` regardless of the position of the `uploads` query argument in the URL (meaning setting  the `request_headers` to `{'uploads':''}` doesn't work either.

Would it be a good idea to have some sort of ordered `extra_args` argument or something to be able to provide the extra query parameters needed in some methods.

http://docs.aws.amazon.com/AmazonS3/latest/API/mpUploadUploadPart.html gets even worse...

soulseekah:
Here's something that is dirty and works for me at the moment:

Cykooz:
I agree - for me it would be very useful.

PR #3322, opened 2015-08-28
Support query params in S3 generate_url_sigv4

zachmullen:
This allows presigning of multipart requests, which require
certain query parameters to be specified and signed.


zachmullen:
Fixes #2852

zachmullen:
Can someone take a look at this? It's a pretty small change and once it's merged I can stop using a hacky workaround :)

PR #3344, opened 2015-09-22
Add GovCloud to Redshift

kyleknap:
Documented here: http://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-endpoints.html

cc @jamesls @mtdowling @rayluo


jamesls:
:shipit:

PR #3348, opened 2015-09-23
Fix SQSConnection.send_message_batch parameter names.

### LINKED ISSUE #3335: boto.sqs.connection.send_message_batch does not work when message attributes are provided
If you try to send a batch of messages with boto.sqs.connection.send_message_batch and have attributes set on the batch items, you will get an error message that the message attribute 'X' must contain non-empty message attribute value.

this is because the string "Value" is missing from the p_name key for each of the cases.

for example: p_name = '%s.%i.DataType' % (base, j + 1)
is supposed to be: p_name = '%s.%i.Value.DataType' % (base, j + 1)

janslow:
If you try to call `Queue.write_batch()` with any message attributes, SQS returns an error and fails to enqueue the message. It returns the following error:

```python
{
  'error_message': u"The message attribute 'PARAMETER_NAME' must contain non-empty message attribute value.",
  'error_code': u'InvalidParameterValue',
  'id': u'0',
  'sender_fault': u'true'
}
```

This is caused by `SQSConnection.send_message_batch()` using the wrong parameter names for message attributes values, so AWS thinks no value/data type has been sent.

See http://queue.amazonaws.com/doc/2012-11-05/QueueService.wsdl for the schema.

Fixes #3335.


janslow:
(I'll fix the tests now)

jamesls:
Thanks for the pull request, looks good to me.

PR #3357, opened 2015-10-03
Decode binary messages

svrana:
Do not attempt to utf-8 decode binary messages.

This made an upgrade from 2.20 to 2.38 a little eventful.


svrana:
Ooops, I'll fix the python 3 test failures and submit a pr against master.

PR #3358, opened 2015-10-03
Fix binary message decoding

svrana:
If utf-8 decoding of base64 message fails, return base64
message.


svrana:
This is pretty ugly, but couldn't figure out a better way to get it to work for python3 without just trying the decode.

PR #3363, opened 2015-10-07
Use UserList in boto ResultSet instead of list, UserList plays nicer …

darjus-amzn:
…with subclassing and fixes test failures for Jython


darjus-amzn:
Sorry about that, obviously, needs more work.

PR #3364, opened 2015-10-07
Use UserList in boto ResultSet instead of list, UserList plays nicer …

darjus-amzn:
…with subclassing and fixes test failures for Jython
Sorry about the botched first attempt.

### DIFF
--- a/boto/resultset.py
+++ b/boto/resultset.py
@@ -19,10 +19,13 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
 # IN THE SOFTWARE.
 
+from collections import Sequence
+from boto.vendored.six.moves import UserList
+
 from boto.s3.user import User
 
 
-class ResultSet(list):
+class ResultSet(UserList):
     """
     The ResultSet is used to pass results back from the Amazon services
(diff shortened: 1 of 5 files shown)


darjus-amzn:
Hey Guys,

Any feedback would be appreciated.

boto/resultset.py
```diff
@@ -19,10 +19,13 @@
...
 # IN THE SOFTWARE.
 
+from collections import Sequence
+from boto.vendored.six.moves import UserList
+
 from boto.s3.user import User
 
 
-class ResultSet(list):
+class ResultSet(UserList):
```
mtdowling: Are you subclassing from `UserList` instead of `list` to satisfy some constraint of Jython?

My main concern here is that this is a subtle breaking change to the API because `UserList` is not a subclass of `list`. This has the unfortunate side effect of breaking any user that is using `isinstance` checks to see if a `ResultSet` is a `list`.

darjus: Yeah, that's a valid concern on the API. I'll see what i can do from Jython side, but I'd like to keep it open for a bit -  maybe someone else will chime in.

darjus:
Got it sorted in Jython instead. Thanks for looking at it!

Issue #3298, opened 2015-08-13
Support AWS_SESSION_TOKEN environment variable


copumpkin:
According to http://blogs.aws.amazon.com/security/post/Tx3D6U6WSFGOK2H/A-New-and-Standardized-Way-to-Manage-Credentials-in-the-AWS-SDKs (and my experience), I can use temporary credentials with `aws-cli` by setting the `AWS_SESSION_TOKEN` environment variable. Boto seems to ignore that variable, even though it picks it up just fine from the metadata service on an EC2 instance.

Note that this seems slightly different from #2988 which (as far as I can tell) has to do with the on-disk configuration format changing.

stan3:
Also the logic currently only uses AWS_SECURITY_TOKEN if AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are not set which doesn't work for me and is not how boto3 works:

    elif ((security_token_name is not None) and
        (access_key is None) and (secret_key is None)):

DanyC97:
is this still open or can be closed?

diasjorge:
I'm having the same problem. I don't have the credentials to the account and rely on using assume role.
Boto makes an assumption here: https://github.com/boto/boto/blob/d9e5cfe900e1a58717e393c76a6e3580305f217a/boto/provider.py#L347-L350 that goes against the recommended way of working from amazon

CrossfireCurt:
I'm also having this same issue.

ayqazi:
Just ran into this issue using Ansible with  boto 2.39 .

If I run Ansible/boto on an EC2 instance with a role, and do not set any credentials in config files or the environment, the libraries generate credentials for the role of the instance and work just fine.

However, if I want to assume a role on a customer's account from my laptop, so I can run Ansible against it, I use `aws sts assume-role` and fill in my environment vars as expected. I can then run aws CLI commands fine, but the Ansible EC2 dynamic inv script fails and I get:
```
ERROR: "Authentication error retrieving ec2 inventory.
 - No AWS_ACCESS_KEY_ID or AWS_SECRET_ACCESS_KEY environment vars found
 - Boto configs found at '~/.aws/credentials', but the credentials contained may not be correct", while: getting EC2 instances
```

FilBot3:
I am also getting the following errors that ayqazi's comment:  https://github.com/boto/boto/issues/3298#issuecomment-191313622 is showing when using Boto with Ansible, or by hand.

FilBot3:
[...]

PR #3385, opened 2015-11-05
Fix aws security token

stan3:
fix issues mentioned in #3298


kapilt:
this solves a couple of headaches we're having generating sts credentials that work across tooling.

brandond:
Why oh why is this still open?

PR #3390, opened 2015-11-12
Add cryptography implementtion for Distribution._sign_string

youknowone:
The cryptograph implementation is quite faster than the rsa implementation.
So the strategy is:

- Try to search cryptography and put it to _sign_string
- Try to search rsa and put it to _sign_string
- Or raise exception

Implemented by @kkung as monkey patch and ported it to boto.


youknowone:
The reason this test is failing is not for code or cryptography but for version of pypy in travis.

PR #3418, opened 2015-12-06
Don't generate auth in url for public files

wsuchy:
Using hmac-v4-s3 authentication causes all URLs to be presigned. Actually we don't need this when generating URL for public bucket.  I have added AWS_QUERYSTRING_AUTH flag support in this scenario.


stefanfoulis:
the failing tests don't seem to be related to this pull request

stefanfoulis:
@wsuchy if you merge in my commit that adds tests from https://github.com/boto/boto/pull/3470, I'll close my pull-request again and we can continue with yours.

PR #3422, opened 2015-12-09
Add deprecation notice to emr methods

kyleknap:
Updated docs to be more in-line with EMR API docs about method deprecation: http://docs.aws.amazon.com/ElasticMapReduce/latest/API/API_DescribeJobFlows.html

cc @jamesls @mtdowling @rayluo @JordonPhillips


mtdowling:
:ship:

Issue #3402, opened 2015-11-23
No support for multiple s3 lifecycle transitions


ryansydnor:

ryansydnor:
This is addressed by https://github.com/boto/boto/pull/3431

anonymuse:
Nice, I'd like to see this implemented. I'll take a look at the related commits.

anonymuse:
Now that #3431 is back working in the CI, this looks gtg.

jamesls:
Fixed via https://github.com/boto/boto/pull/3431

PR #3431, opened 2015-12-17
Allow s3 bucket lifecycle policies with multiple transitions

ryansydnor:
Addresses issue #3402

### DIFF
--- a/tests/unit/s3/test_lifecycle.py
+++ b/tests/unit/s3/test_lifecycle.py
@@ -50,27 +50,103 @@ class TestS3LifeCycle(AWSMockServiceTestCase):
             <Status>Disabled</Status>
             <Transition>
               <Date>2012-12-31T00:00:000Z</Date>
-              <StorageClass>GLACIER</StorageClass>
+              <StorageClass>STANDARD_IA</StorageClass>
+            </Transition>
+            <Expiration>
+              <Date>2012-12-31T00:00:000Z</Date>
+            </Expiration>
+          </Rule>
+          <Rule>
+            <ID>multiple-transitions</ID>
+            <Prefix></Prefix>
+            <Status>Enabled</Status>
+            <Transition>
+                <Days>30</Days>
+                <StorageClass>STANDARD_IA</StorageClass>
+            </Transition>
+            <Transition>
+                <Days>90</Days>
+                <StorageClass>GLACIER</StorageClass>
             </Transition>
           </Rule>
         </LifecycleConfiguration>
         """
 
-    def test_parse_lifecycle_response(self):
+    def _get_bucket_lifecycle_config(self):
         self.set_http_response(status_code=200)
         bucket = Bucket(self.service_connection, 'mybucket')
-        response = bucket.get_lifecycle_config()
-        self.assertEqual(len(response), 2)
-        rule = response[0]
+        return bucket.get_lifecycle_config()
+
+    def test_lifecycle_response_contains_all_rules(self):
+        self.assertEqual(len(self._get_bucket_lifecycle_config()), 3)
+
+    def test_parse_lifecycle_id(self):
+        rule = self._get_bucket_lifecycle_config()[0]
         self.assertEqual(rule.id, 'rule-1')
+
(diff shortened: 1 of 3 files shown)


copumpkin:
Have you looked at the travis failure? Is it spurious or legitimate?

ryansydnor:
Currently unclear. I see at least two other PRs (https://github.com/boto/boto/pull/3429 & https://github.com/boto/boto/pull/3418) failing with exactly the same issues on Python 3.3/3.4.

copumpkin:
Cool, assuming the Travis thing isn't a thing, :+1: from me. How do we get attention to this from project collaborators?

ryansydnor:
@jamesls - How can I help move this along? 

There is an upstream PR dependent on this (https://github.com/ansible/ansible-modules-extras/pull/1393) and people expressing interest in the feature request (https://github.com/boto/boto/issues/3402).

anonymuse:
Looks like Travis is back on track :+1: 

#3402 seems gtg.

ryansydnor:
@jamesls - anything I can do to help move this along?

jamesls:
Taking a look, I just want to double check there are no backwards compatibility concerns.

ryansydnor:
Great! I tried to make the interface backwards compatible. Thanks for the double check on logical errors :)

tests/unit/s3/test_lifecycle.py
```diff
@@ -50,27 +50,90 @@ def default_body(self):
...
+    def test_parse_lifecycle_disabled(self):
+        rule = self._get_bucket_lifecycle_config()[1]
+        self.assertEqual(rule.status, 'Disabled')
+
+    def test_parse_expiration_days(self):
+        rule = self._get_bucket_lifecycle_config()[0]
         self.assertEqual(rule.expiration.days, 365)
-        self.assertIsNone(rule.expiration.date)
-        transition = rule.transition
-        self.assertEqual(transition.days, 30)
```
jamesls: Seems like this behavior would no longer work, correct?  Because `transition` is now a list, instead of `rule.transition.days`, you'd have to say `rule.transition[0].days`.

jamesls:
Seems like we'd have to somehow update the `Transitions` class to have properties that would correspond to the first `Transition` to keep backwards compatibility.

What other SDKs (including boto3) ended up doing, is just introducing a new method in the S3 client that parsed the correct way (with transitions as a list) and updated the existing method's documentation to say it's deprecated and that you should use the newer method.  [get_bucket_lifecycle](http://boto3.readthedocs.org/en/latest/reference/services/s3.html#S3.Client.get_bucket_lifecycle) vs. [get_bucket_lifecycle_configuration](http://boto3.readthedocs.org/en/latest/reference/services/s3.html#S3.Client.get_bucket_lifecycle_configuration).  Might be worth considering.

tests/unit/s3/test_lifecycle.py
ryansydnor: Correct (see test_parse_transition_days test below to confirm).

ryansydnor:
@jamesls - you're right. This breaks backwards compatibility in that case. 

I can also imagine an alternative approach via the @property decorator. Include a "getter" for date, days, and storage_class. Something like:

```python
def _first_or_default(self, property, default=None):
    for transition in self:
        return transition.get(property, default)
    return default

@property
def date(self):
    return self._first_or_default('date')

@property
def days(self):
    return self._first_or_default('days')

@property
def storage_class(self):
    return self._first_or_default('storage_class')
```

In this model, we can still mention the API to directly access these properties is deprecated in the docs.

What do you think?

ryansydnor:
@jamesls - this should now be 100% backwards compatible. Let me know what else I can do to help push it along.

tests/unit/s3/test_lifecycle.py
ryansydnor: Fixed in the next iteration of the PR.

jamesls:
Looks good.  Your suggested approach seems reasonable.  Thanks for updating.

ryansydnor:
Thanks :)

PR #3432, opened 2015-12-18
Update docs for deprecate_activity_type

t0m:
Set to match https://docs.aws.amazon.com/amazonswf/latest/apireference/API_DeprecateActivityType.html


jamesls:
Thanks!

Issue #3427, opened 2015-12-16
[Feature Request] - Add support for encrypted EBS boot volumes


ryansydnor:
Encrypted EBS boot volumes were recently [announced](https://aws.amazon.com/blogs/aws/new-encrypted-ebs-boot-volumes/).

I could imagine [copy_image](https://github.com/boto/boto/blob/d9e5cfe900e1a58717e393c76a6e3580305f217a/boto/ec2/connection.py#L4408) taking in the two new parameters 'encrypted' and 'kmsKeyID'. And, of course, propagating those through to [CopyImage](https://github.com/boto/boto/blob/d9e5cfe900e1a58717e393c76a6e3580305f217a/boto/ec2/image.py#L435)

jamesls:
Fixed in https://github.com/boto/boto/pull/3438

PR #3438, opened 2015-12-20
Add encryption capability to AMI copy

ryansydnor:
This addresses https://github.com/boto/boto/issues/3427

### DIFF
--- a/boto/ec2/connection.py
+++ b/boto/ec2/connection.py
@@ -4405,7 +4405,8 @@ class EC2Connection(AWSQueryConnection):
         return self.get_list('DescribeInstanceTypes', params, [('item', InstanceType)], verb='POST')
 
     def copy_image(self, source_region, source_image_id, name=None,
-                   description=None, client_token=None, dry_run=False):
+                   description=None, client_token=None, dry_run=False,
+                   encrypted=None, kms_key_id=None):
         """
         :type dry_run: bool
         :param dry_run: Set to True if the operation should not actually run.
@@ -4422,6 +4423,10 @@ class EC2Connection(AWSQueryConnection):
             params['Description'] = description
         if client_token is not None:
             params['ClientToken'] = client_token
+        if encrypted is not None:
+            params['Encrypted'] = 'true' if encrypted else 'false'
+        if kms_key_id is not None:
+            params['KmsKeyId'] = kms_key_id
         if dry_run:
             params['DryRun'] = 'true'
         return self.get_object('CopyImage', params, CopyImage,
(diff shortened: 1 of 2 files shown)


boto/ec2/connection.py
```diff
@@ -4406,7 +4406,8 @@ def get_all_instance_types(self):
         return self.get_list('DescribeInstanceTypes', params, [('item', InstanceType)], verb='POST')
 
     def copy_image(self, source_region, source_image_id, name=None,
-                   description=None, client_token=None, dry_run=False):
+                   description=None, client_token=None, encrypted=None,
```
jamesls: I think we'd need to add `kms_key_id` and `encrypted` after `dry_run` to preserve backwards compatibility for people using positional args.

ryansydnor:
@jamesls - agree with your comment about potentially breaking backwards compatibility by shuffling parameter positions. fixed!

i also took the opportunity to only include KmsKeyId if Encrypted is also included :)

ryansydnor:
@jamesls - I don't believe this PR is pending-feedback any longer.

There is an upstream PR that depends on this being merged (https://github.com/ansible/ansible-modules-extras/pull/1409).

How can I help move this along?

ryansydnor:
@jamesls - we are trying to merge the upstream PR. What do we need to do to move this along?

jamesls:
Taking a look.

boto/ec2/connection.py
```diff
@@ -4423,6 +4424,11 @@ def copy_image(self, source_region, source_image_id, name=None,
             params['Description'] = description
         if client_token is not None:
             params['ClientToken'] = client_token
+        if encrypted is not None:
+            params['Encrypted'] = 'true' if encrypted else 'false'
+
+            if kms_key_id is not None:
```
jamesls: I think we should add `kms_key_id` if the user provides it, regardless of whether or not they specify `encrypted`.  If the user mistakenly adds `kms_key_id` without `encrypted` then we silently ignore `kms_key_id` and the user may not know anything is wrong.  If we always add the `kms_key_id` they'll at least get an error from EC2 saying that you have to include `encrypted` if you're including `kms_key_id`.

ryansydnor: Ok. I used the same logic as create_volume above. Wanted to prevent the user from shooting themselves in the foot, but agree that this just gives them a silencer!

Updated the PR and rebased off the develop branch. Please take a look!

ryansydnor:
@jamesls - updated and rebased off develop based on your comments :)

jamesls:
Looks good, thanks for updating.

Issue #3263, opened 2015-07-09
Problem importing boto in Python 3 running under Google Cloud


sbneto:
Importing the module fails with the following message:
```
>>> import boto
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.4/dist-packages/boto/__init__.py", line 1216, in <module>
    boto.plugin.load_plugins(config)
  File "/usr/local/lib/python3.4/dist-packages/boto/plugin.py", line 93, in load_plugins
    _import_module(file)
  File "/usr/local/lib/python3.4/dist-packages/boto/plugin.py", line 75, in _import_module
    return imp.load_module(name, file, filename, data)
  File "/usr/lib/python3.4/imp.py", line 235, in load_module
    return load_source(name, filename, file)
  File "/usr/lib/python3.4/imp.py", line 171, in load_source
    module = methods.load()
  File "/usr/share/google/boto/boto_plugins/compute_auth.py", line 64
    except (urllib2.URLError, urllib2.HTTPError, IOError), e:
```
After changing the line
```
except (urllib2.URLError, urllib2.HTTPError, IOError), e:
```
to 
```
except (urllib2.URLError, urllib2.HTTPError, IOError) as e:
```
I got the error:
```
>>> import boto
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.4/dist-packages/boto/__init__.py", line 1216, in <module>
    boto.plugin.load_plugins(config)
  File "/usr/local/lib/python3.4/dist-packages/boto/plugin.py", line 93, in load_plugins
    _import_module(file)
  File "/usr/local/lib/python3.4/dist-packages/boto/plugin.py", line 75, in _import_module
    return imp.load_module(name, file, filename, data)
  File "/usr/lib/python3.4/imp.py", line 235, in load_module
    return load_source(name, filename, file)
  File "/usr/lib/python3.4/imp.py", line 171, in load_source
    module = methods.load()
  File "/usr/share/google/boto/boto_plugins/compute_auth.py", line 18, in <module>
    import urllib2
ImportError: No module named 'urllib2'
```
So I simply imported urllib3 as urllib2:
```
import urllib3 as urllib2
```

This seemed to solve the problem and I am now able to import boto. I haven't fully tested it to verify everything is working though (although I don't see why it shouldn't). For the record, my Python version is 3.4.2 and boto version is 2.38.0

Regards,

Sam

[...]

PR #3439, opened 2015-12-22
Fix travis CI builds for PY3

jamesls:
Tracking issue here: https://github.com/travis-ci/travis-ci/issues/5246.
What's happening is that the global boto config on the travis CI
systems is telling boto to load a plugin that is not compatible with
python3.

In the meantime, our workaround is to set the BOTO_CONFIG to some bogus
value so we don't load the global config.

Verified we get a clean run through travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ python:
   - "3.3"
   - "3.4"
   - "pypy"
+env:
+  - BOTO_CONFIG=/tmp/nowhere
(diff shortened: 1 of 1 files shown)


ryansydnor:
+1

Thanks for helping track down the issue so quickly!

rayluo:
Awesome! :ship:

jamesls:
Fixes #3263

PR #3441, opened 2015-12-22
Get all unit tests passing on python3.

jamesls:
This removed the py3 whitelist.  Now all unit tests pass on python 3.

I did have to make one change to the auth class used by mturk, but it's a pretty mechanical change.

cc @kyleknap @mtdowling @rayluo @JordonPhillips


rayluo:
:ship:

Issue #3349, opened 2015-09-23
DynamoDB Global index capacity read from AWS is not used


mballou11:
The throughput value for a global secondary index read from the actual AWS table is never read, and always defaults to 5.

The following code needs to be added to table.py inside _introspect_all_indexes.
            if 'ProvisionedThroughput' in field:
                raw_throughput = field['ProvisionedThroughput']
                kwargs['throughput'] = {}
                kwargs['throughput']['read'] = int(raw_throughput['ReadCapacityUnits'])
                kwargs['throughput']['write']  = int(raw_throughput['WriteCapacityUnits'])

ferrants:
I am running into this as well. I've been able to track the issue to these lines:
https://github.com/boto/boto/blob/master/boto/dynamodb2/fields.py#L222
https://github.com/boto/boto/blob/master/boto/dynamodb2/table.py#L110

@mballou11 I implemented your suggestion here. https://github.com/boto/boto/pull/3447

PR #3448, opened 2016-01-04
fix for throughput not returning correctly

ferrants:
Issue: https://github.com/boto/boto/issues/3349

calling `table.global_indexes[0].throughput` always returns

```
{'read': 5, 'write': 5}
```

same results when you look at `index.schema()`:

```
{u'BatchServerAliasIndex': {'IndexName': u'BatchServerAliasIndex', 'Projection': {'ProjectionType': 'INCLUDE', 'NonKeyAttributes': [u'log_filename', u'workflow']}, 'ProvisionedThroughput': {'WriteCapacityUnits': 5, 'ReadCapacityUnits': 5}, 'KeySchema': [{'KeyType': 'HASH', 'AttributeName': u'batch_id'}, {'KeyType': 'RANGE', 'AttributeName': u'server_alias'}]}}
```

I've been able to track the issue to these lines:
https://github.com/boto/boto/blob/develop/boto/dynamodb2/fields.py#L222
https://github.com/boto/boto/blob/develop/boto/dynamodb2/table.py#L110

### DIFF
--- a/boto/dynamodb2/table.py
+++ b/boto/dynamodb2/table.py
@@ -307,6 +307,12 @@ class Table(object):
                     field['Projection']['ProjectionType']
                 )
 
+            if 'ProvisionedThroughput' in field:
+                raw_throughput = field['ProvisionedThroughput']
+                kwargs['throughput'] = {}
+                kwargs['throughput']['read'] = int(raw_throughput['ReadCapacityUnits'])
+                kwargs['throughput']['write'] = int(raw_throughput['WriteCapacityUnits'])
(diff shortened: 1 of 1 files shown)


kyleknap:
Could you please add tests for your fix?

ferrants:
There are already tests that should catch this. They don't because they don't actually pull from the AWS servers, they use mocks. I'll look into another test

ferrants:
I use boto3 now, don't care, closing, leaving branch.

PR #3451, opened 2016-01-05
List instances improvements

jimbrowne:
I still use list_instances as a quick alternative to the AWS CLI.  These commits:

* add support for 19 more instance fields (including the IAM instance profile)
* changes the default header set removing "Hostname" and adding the "Name" tag to match the AWS console
* sets a default column width for fields when not specified
* PEP8 cleanups

No tests included as there are no existing tests for the command line tools.

### DIFF
--- a/bin/list_instances
+++ b/bin/list_instances
@@ -57,12 +57,20 @@ def main():
 
     # Create format string
     format_string = ""
-    for h in headers:
-        if h.startswith('T:'):
+    for header in headers:
+        if header.startswith('T:'):
             format_string += "%%-%ds" % HEADERS['T:']['length']
         else:
-            format_string += "%%-%ds" % HEADERS[h]['length']
+            length = 0
 
+            if 'length' in HEADERS[header]:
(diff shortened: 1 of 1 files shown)


jamesls:
Not opposed to the changes provided this is backwards compatible.

Also wondering, maybe it's worth adding a simplified "list-instances" command to the AWS CLI?

jimbrowne:
The only change in behavior which might be considered backwards incompatible is changing the default header set to remove "Hostname" and add the "Name" tag to match what the AWS console does.

jamesls:
I can see the motivation for matching the console output, but could we add name as an additional column at the end instead of replacing hostname?  Seems like a reasonable compromise between backwards compatibility and being similar to the console.

jimbrowne:
I'll migrate the column change to a command line flag; ETA late February.

jimbrowne:
I have brought back the Hostname column and left the T:Name column as an addition as suggested.

I also changed the ID column length to accommodate the new-style 17 character instance identifiers.

PR #3459, opened 2016-01-11
Fixed to make previous commit python3 safe

kopertop:
Sorry I hadn't updated my local checkout of boto to make the first commit through a pull-request.

This fix actually fixes my previous commit directly to boto/develop, which is intended to fix in some scenarios where Lists seem to come back from AWS as simple strings.

### DIFF
--- a/boto/dynamodb/types.py
+++ b/boto/dynamodb/types.py
@@ -333,7 +333,7 @@ class Dynamizer(object):
         the appropriate python type.
 
         """
-        if len(attr) > 1 or not attr or isinstance(attr, basestring):
+        if len(attr) > 1 or not attr or is_str(attr):
             return attr
         dynamodb_type = list(attr.keys())[0]
         if dynamodb_type.lower() == dynamodb_type:
--- a/tests/unit/dynamodb/test_types.py
+++ b/tests/unit/dynamodb/test_types.py
@@ -23,7 +23,7 @@
 from decimal import Decimal
 from tests.compat import unittest
 
-from boto.compat import six
+from boto.compat import six, json
 from boto.dynamodb import types
 from boto.dynamodb.exceptions import DynamoDBNumberError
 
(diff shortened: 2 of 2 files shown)


kyleknap:
Can we get a test added to this PR for the previous commit that you added that tests the issue you fixed?

kyleknap:
I was just thinking about this. Is this change: https://github.com/boto/boto/commit/059f525b8731556f4878cc46a7712f7c92f91940 even needed? Could you provide an example of how this edge case is entered? I am having trouble thinking of a way to reach this edge case.

kopertop:
Yes, I ran across a very weird edge case that caused problems, let me see
if I can get an example together and build a test where you can see without
this check it fails.

On Mon, Jan 11, 2016 at 2:02 PM, Kyle Knapp <notifications@github.com>
wrote:

> I was just thinking about this. Is this change: 059f525
> <https://github.com/boto/boto/commit/059f525b8731556f4878cc46a7712f7c92f91940>
> even needed? Could you provide an example of how this edge case is entered?
> I am having trouble thinking of a way to reach this edge case.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/boto/boto/pull/3459#issuecomment-170653832>.
>



-- 
Chris Moyer

kopertop:
I added a test which shows this failing, the problem is only when you do a full decode via the ```object_hook``` attribute of ```json.loads```

```
        dynamizer = types.Dynamizer()
        doc = '{"__type__":{"S":"Story"},"company_tickers":{"SS":["NASDAQ-TSLA","NYSE-F","NYSE-GM"]},"modified_at":{"N":"1452525162"},"created_at":{"N":"1452525162"},"version":{"N":"1"},"categories":{"SS":["AUTOMTVE","LTRTR","MANUFCTU","PN","PRHYPE","TAXE","TJ","TL"]},"provider_categories":{"L":[{"S":"F"},{"S":"GM"},{"S":"TSLA"}]},"received_at":{"S":"2016-01-11T11:26:31Z"}}'
        output_doc = {'provider_categories': ['F', 'GM', 'TSLA'], '__type__': 'Story', 'company_tickers': set(['NASDAQ-TSLA', 'NYSE-GM', 'NYSE-F']), 'modified_at': Decimal('1452525162'), 'version': Decimal('1'), 'received_at': '2016-01-11T11:26:31Z', 'created_at': Decimal('1452525162'), 'categories': set(['LTRTR', 'TAXE', 'MANUFCTU', 'TL', 'TJ', 'AUTOMTVE', 'PRHYPE', 'PN'])}
        self.assertEqual(json.loads(doc, object_hook=dynamizer.decode), output_doc)
```

Without this patch, the above tests fails with:

```
self = <boto.dynamodb.types.Dynamizer object at 0x10d2e29d0>, attr = ['F', 'GM', 'TSLA']

    def _decode_l(self, attr):
>       return [self.decode(i) for i in attr]

boto/dynamodb/types.py:377: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <boto.dynamodb.types.Dynamizer object at 0x10d2e29d0>, attr = 'F'

    def decode(self, attr):
        """
            Takes the format returned by DynamoDB and constructs
            the appropriate python type.
    
            """
        if len(attr) > 1 or not attr:
            return attr
>       dynamodb_type = list(attr.keys())[0]
E       AttributeError: 'str' object has no attribute 'keys'

boto/dynamodb/types.py:338: AttributeError
```

kyleknap:
Thanks @kopertop. That's an interesting use case. Have not seen it used with ``load`` or ``dump``. I was curious into why this was happening and it is because the ``loads`` function will actually decode the nested dictionaries first as opposed to how the dynamizer decodes out objects first and recurses in. So ``json.loads`` will decode ``{"S":"F"}``,``{"S":"GM"}``, and ``{"S":"TSLA"}`` before decoding ``"provider_categories":{"L":[{"S":"F"},{"S":"GM"},{"S":"TSLA"}]}``. As a result when ``json.loads`` gets to the ``provider_categories`` attribute the value will be ``"provider_categories":{"L":["F","M","TSLA"]}`` and that will cause the dynamizer to fail because that is not valid syntax. Now the check makes sense. Merging.

PR #3471, opened 2016-01-22
Fix whitespace-collapsing for canonical headers

tipabu:
This is related to #2142 -- presumably the docs were updated, as http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html now reads:

> **Original headers**
>```
>   Host:iam.amazonaws.com\n
>   Content-Type:application/x-www-form-urlencoded; charset=utf-8\n
>   My-header1:    a   b   c  \n
>   X-Amz-Date:20150830T123600Z\n
>   My-Header2:    "a   b   c"  \n
>```
> **Canonical form**
>```
>   content-type:application/x-www-form-urlencoded; charset=utf-8\n
>   host:iam.amazonaws.com\n
>   my-header1:a b c\n
>   my-header2:"a b c"\n
>   x-amz-date:20150830T123600Z\n
>```

Tested against s3-us-west-2, the old behavior (preserve spaces within quotes) resulted in SignatureDoesNotMatch errors, while the old-old behavior (convert sequential spaces to a single space, regardless of quote marks) was accepted.

### DIFF
--- a/boto/auth.py
+++ b/boto/auth.py
@@ -374,10 +374,7 @@ class HmacAuthV4Handler(AuthHandler, HmacKeys):
         for header in headers_to_sign:
             c_name = header.lower().strip()
             raw_value = str(headers_to_sign[header])
-            if '"' in raw_value:
-                c_value = raw_value.strip()
-            else:
-                c_value = ' '.join(raw_value.strip().split())
+            c_value = ' '.join(raw_value.strip().split())
             canonical.append('%s:%s' % (c_name, c_value))
         return '\n'.join(sorted(canonical))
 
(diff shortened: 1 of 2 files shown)


kyleknap:
Interesting. I wonder why that if statement was there in the first place. Will need to do research to make sure we do not break anything.

tipabu:
Anything else I can/should do here? Wayback Machine [confirms the old docs](https://web.archive.org/web/20140328141813/http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html) from #2142, which persisted as far as [September 2015](https://web.archive.org/web/20150908074849/http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html). But by [January 2016](https://web.archive.org/web/20160126120845/http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html) they changed and those changes are still reflected in the current docs.

PR #3478, opened 2016-01-27
fix documentation for get_shard_iterator

ferozed:
The documentation of get_shard_iterator is not correct. It implies that you can directly use the return value of get_shard_iterator() in a get_records call.

If you do that, you will get an error from kinesis api. That is because the output of get_shard_iterator is actually a dictionary, and not the real shard. 

The attached pull request fixes the documentation.


PauloMigAlmeida:
Good mate :+1:

ferozed:
hi! When will this be merged?

kyleknap:
Looks good.

PR #3482, opened 2016-01-29
Fix non-pr travis builds of master

kyleknap:
The environment variable is "false" if it is a PR. However that equates to
true as it is interpreted as a string

cc @jamesls @mtdowling @rayluo @JordonPhillips


mtdowling:
:ship:

PR #3483, opened 2016-01-30
Decode request bodies before passing them to json

gholms:
AWSAuthConnection._mexe encodes request bodies as UTF-8, mutating the
original request object in the process.  This breaks kinesis's unit
tests on at least python 3.4 and 3.5 on Fedora because those unit tests
call json.loads, which expects str objects rather than the bytes objects
that _mexe converted them to.

Frankly, I'm a little confused as to how those tests manage to pass on
travis-ci.  If they somehow on this branch fail I guess I might find out why.

This commit makes the test cases decode request bodies before feeding
them to json.loads.


jamesls:
:shipit:  Looks good to me.

PR #3485, opened 2016-02-03
Python 3 forward compatibility for mturk

NickShaffner:
Python 3 map() forward compatability as well as an explicit float to int conversion for pagination.


NickShaffner:
hrmm, no future.builtins.iterators in botos packages, need to learn more about where these go.

PR #3488, opened 2016-02-05
Un-quoting multipart uploads and versioned objects

mtdowling:
This commit updates Boto to decode multipart upload markers and
versioned bucket markers to prevent infinite loops when listing.

Closes #2948 

@kyleknap @jamesls @rayluo @JordonPhillips


kyleknap:
Thanks :ship: once the travis build passes.

PR #3497, opened 2016-02-17
Fixed issue with AttributeError exception being raised

doihaveto:
AttributeError exception being raised when self.curItem is None


kyleknap:
Can you please add a test for this fix?

doihaveto:
I'm not sure how I would go about doing that. The `AttributeError` issue only occurs when specific parent ASINs are passed to the `item_lookup` method, with `ResponseGroup='Variations'` parameter passed as well. Example ASIN: B0018OMIMK.

PR #3498, opened 2016-02-19
Fix AWS credentials file loading for Python 3

### LINKED ISSUE #3474: Config not correctly parsing aws credentials file in python 3.5
Looks like there is some issue with how the custom `Config` class is overriding the `ConfigParser` class in python 3.5.1 on Mac OSX 10.10.5. I am not able to read my aws credentials from the default `~/.aws/credentials` file if I use the boto `Config` class, but am able to if I use the `boto.compat.ConfigParser` class.

```
Python 3.5.1 (v3.5.1:37a07cee5969, Dec  5 2015, 21:12:44)
[GCC 4.2.1 (Apple Inc. build 5666) (dot 3)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from boto.compat import ConfigParser
>>> c = ConfigParser()
>>> c.read("/Users/asgoel/.aws/credentials")
['/Users/asgoel/.aws/credentials']
>>> c.get("default", "aws_access_key_id")
RETURNS_AWS_ACCESS_KEY
>>> from boto.pyami.config import Config
>>> c1 = Config()
c1.read("/Users/ashu/.aws/credentials")
['/Users/ashu/.aws/credentials']
>>> print(c1.get("default", "aws_access_key_id"))
None
```

Makes running applications locally a little bit annoying but not a huge issue, just something I thought I would point out to see if anyone is having similar problems. I couple of others within my organization are facing the same problem.

tsavola:
Fixes #3474

### DIFF
--- a/boto/pyami/config.py
+++ b/boto/pyami/config.py
@@ -26,7 +26,7 @@ import warnings
 
 import boto
 
-from boto.compat import expanduser, ConfigParser, StringIO
+from boto.compat import expanduser, ConfigParser, NoOptionError, NoSectionError, StringIO
 
 
 # By default we use two locations for the boto configurations,
@@ -128,21 +128,21 @@ class Config(object):
     def get_instance(self, name, default=None):
         try:
             val = self.get('Instance', name)
-        except:
+        except (NoOptionError, NoSectionError):
             val = default
         return val
 
     def get_user(self, name, default=None):
         try:
             val = self.get('User', name)
-        except:
+        except (NoOptionError, NoSectionError):
             val = default
         return val
 
     def getint_user(self, name, default=0):
         try:
             val = self.getint('User', name)
-        except:
+        except (NoOptionError, NoSectionError):
             val = default
         return val
 
@@ -156,7 +156,7 @@ class Config(object):
             return default
         try:
             return impl.get(section, name)
-        except:
+        except (NoOptionError, NoSectionError):
             return default
 
     def getint(self, section, name, default=0):
@@ -166,7 +166,7 @@ class Config(object):
             return int(default)
         try:
             return impl.getint(section, name)
-        except:
+        except (NoOptionError, NoSectionError):
             return int(default)
 
     def getfloat(self, section, name, default=0.0):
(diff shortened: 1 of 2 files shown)


kyleknap:
I have not looked too deep into the pull request but to start can you please add a test for this fix?

blacked:
+1 for that PR. 
2.49.0 boto.config doesn't work with Python3.

marco-hoyer:
@tsavola do you need any support here? Would be great to have this fixed.

tsavola:
I replaced the config-fix branch contents with a refactored version (since pull requests don't seem to support updating the branch). It has some additional fixes from the version I use in production.

The original version is here: https://github.com/ninchat/boto/commits/config-fix1

boto/pyami/config.py
```diff
@@ -70,6 +68,20 @@ def __init__(self, path=None, fp=None, do_load=True):
...
+    def __getattr__(self, name):
+        try:
+            impl = self.__dict__['_impl']
+        except KeyError:
+            raise AttributeError(name)
+        return getattr(impl, name)
+
+    def has_option(self, *args, **kwargs):
+        try:
+            impl = self.__dict__['_impl']
```
jamesls: Shouldn't you be able to access `self._impl` directly?  You won't fall back to getattr unless it can't find the attr in the class.

jamesls:
Thanks for the pull request.  The last remaining piece of work is adding tests for this.  We'd like to ensure we don't cause any regressions as well as prevent regressions from happening in the future.

I'm working on tests for this, but otherwise this should be good to merge.  I'm going to make a few small changes.

jamesls:
I took your PR and added a few tests cases that highlight the issue on py3.5.1 along with a few small code changes.  Tracking this update here: https://github.com/boto/boto/pull/3522

boto/pyami/config.py
tsavola: I had a strange pickle-related failure (if I remember correctly) due to missing _impl, but I can't find/reproduce it. It may have been caused by the way I patched the release-version boto with the fixed methods.

PR #3500, opened 2016-02-24
Fix corner case when returning DynamoDB results.

aeturnum:
When reading from an overloaded DynamoDB table, it's possible for AWS to process
no results. As a result, we must check both the results and unprocessed_results
from AWS to avoid returning early and missing unprocessed_results.


kyleknap:
Could you please add a test for this fix and make sure that the tests pass?

PR #3508, opened 2016-03-09
Bugfix: respect max_retry_delay value from config

orlenko:
## Bug:

Current code effectively disregards `max_retry_delay` parameter from config, because it is being retrieved as string.
## Fix:

Use `config.getfloat()` instead of `config.get()` to get the numerical value of the parameter. 
## Background:

In Python comparison rules, any number is less than any string. That's why `max(some_num, some_string)` will always return `some_num`, which is what happens in this case: the random value of sleep time is always chosen over the specified `max_retry_delay`.

Issue: https://github.com/boto/boto/issues/3509

### DIFF
--- a/boto/connection.py
+++ b/boto/connection.py
@@ -920,7 +920,7 @@ class AWSAuthConnection(object):
         while i <= num_retries:
             # Use binary exponential backoff to desynchronize client requests.
             next_sleep = min(random.random() * (2 ** i),
-                             boto.config.get('Boto', 'max_retry_delay', 60))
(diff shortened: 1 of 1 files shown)


gholms:
We have been carrying this patch in Fedora and EPEL for a few months now.  Does it need more review or is it ready to merge?

orlenko:
*bump*

orlenko:
Is there a problem with this PR, or are we simply waiting for it to grow old enough to buy alcohol by itself?

PR #3511, opened 2016-03-10
Correct a grammatical error: "on it's own"

noenglishprofessorbut:
I am no English professor, but "on its own" should not contain an apostrophe. 

For reference, see http://theoatmeal.com/comics/apostrophe


JordonPhillips:
:boat:

PR #3513, opened 2016-03-15
Send correct Host header when using v4 auth with non-default port

tipabu:
Previously, when using a non-default port and v4 authentication, we would send a `Host` header like `<hostname>:<port>:<port>` instead of `<hostname>:<port>`. This would happen because `S3Connection` would explicitly set the host from `self.server_name()` (which includes port) during `make_request` ([code](https://github.com/boto/boto/blob/2.39.0/boto/s3/connection.py#L655)) while `S3HmacAuthV4Handler` would always append the non-standard port to the host (even if it already included a port; [code](https://github.com/boto/boto/blob/2.39.0/boto/auth.py#L592-L597)).

Now, `_CallingFormat`'s `build_host` will only ever return hostnames. This is comparable to what was done in `AWSAuthConnection`'s `new_http_connection` ([code](https://github.com/boto/boto/blob/2.39.0/boto/connection.py#L727-L729)) and seemed

 * more in line with what `build_host` should return (i.e., just a hostname, no port) and
 * less likely to have unintended side effects than a similar change to `server_name`.

Addresses #2623.

### DIFF
--- a/boto/s3/connection.py
+++ b/boto/s3/connection.py
@@ -82,6 +82,10 @@ class _CallingFormat(object):
         return url_base
 
     def build_host(self, server, bucket):
+        # Make sure the server is really just the host, not including
+        # the port number
+        server = boto.utils.parse_host(server)
+
         if bucket == '':
             return server
         else:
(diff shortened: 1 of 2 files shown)


bacongobbler:
Late to the party, but this PR worked for me with minio. Nice patch!

clayg:
@kdaily this one looks good to me!  What do you think?

PR #3515, opened 2016-03-21
Stack name fix

JordonPhillips:
This builds off of #3228, adding an alias from the old name to the new name so that anybody attempting to use the old name isn't broken.

cc @kyleknap @jamesls

### DIFF
--- a/tests/unit/cloudformation/test_connection.py
+++ b/tests/unit/cloudformation/test_connection.py
@@ -444,7 +444,7 @@ class TestCloudFormationDescribeStacks(CloudFormationConnectionBase):
         self.assertEqual(stack.stack_id, 'arn:aws:cfn:us-east-1:1:stack')
         self.assertEqual(stack.stack_status, 'CREATE_COMPLETE')
         self.assertEqual(stack.stack_name, 'MyStack')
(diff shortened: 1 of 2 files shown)


tests/unit/cloudformation/test_connection.py
```diff
@@ -444,7 +444,8 @@ def test_describe_stacks(self):
         self.assertEqual(stack.stack_id, 'arn:aws:cfn:us-east-1:1:stack')
         self.assertEqual(stack.stack_status, 'CREATE_COMPLETE')
         self.assertEqual(stack.stack_name, 'MyStack')
-        self.assertEqual(stack.stack_name_reason, None)
+        self.assertEqual(stack.stack_name_reason, '')
+        self.assertEqual(stack.stack_status_reason, '')
```
jamesls: Can we update the response body in this test to have a StackStatusReason so we can verify we're parsing the expected value?

jamesls:
Looks good, had a small comment on the test.

JordonPhillips:
Updated, the value is now `REASON`

jamesls:
:shipit:

PR #3516, opened 2016-03-22
Make client_token optional in modify_reserved_instances

JordonPhillips:
Previously, we had been serializing "None" instead of omitting the optional parameter.

cc @kyleknap @jamesls


jamesls:
:shipit:

PR #3522, opened 2016-04-01
Fix config parsing on python3.5

jamesls:
This PR pulls in https://github.com/boto/boto/pull/3498 with a few modifications:

* Remove usage of attr access through `__dict__`.
* Add unit tests for fix.

I verified that the added tests fail on python 3.5.1 without this fix.

Fixes https://github.com/boto/boto/issues/3474

cc @kyleknap @JordonPhillips


kyleknap:
:ship:

PR #3527, opened 2016-04-17
[FIX] Verify read permissions on shared path

AvnerCohen:
Bug:
In some system scenarios, it is possible that an .aws file will exists but process owner does not have permission to it (www-data type user). In which case, creds are passed in the constructor.

Fix:
Even when file exists, make sure it is user has read access to it.


AvnerCohen:
Bump. Anyone with any thoughts around that?

Issue #3476, opened 2016-01-27
parametergroup check for allowed_values fails for range with negative numbers


serioushaircut:
In https://github.com/boto/boto/blob/develop/boto/rds/parametergroup.py line 149
```
...
            if self.allowed_values:
                min, max = self.allowed_values.split('-')
                if value < int(min) or value > int(max):
...
```
a range for allowed values is checked.

The split does apparently not quite work when allowed_range features negative values. 
e.g. postgres max_standby_streaming_delay has allowed values '-1-2147483647'
Hence, the split's gonna return ['', '1', '2147483647'], then boom.

Negative ranges are quite common in postgres parameter groups (esp. -1 as the special case like wait forever) btw.
Cheers
SH

PR #3533, opened 2016-04-28
Fix for #3476

robbwagoner:
Several Postgres9.x engine parameters support a range of values, with a negative integer as the minimum. 

For example, from running `aws rds describe-db-parameters --db-parameter-group-name 'default.postgres9.3'`:

```
...
        {
            "ApplyMethod": "pending-reboot",
            "Description": "(ms) Vacuum cost delay in milliseconds, for autovacuum.",
            "DataType": "integer",
            "AllowedValues": "-1-100",
            "Source": "engine-default",
            "IsModifiable": true,
            "ParameterName": "autovacuum_vacuum_cost_delay",
            "ApplyType": "dynamic"
        },
...
```
(`"AllowedValues": "-1-100",`).

This change will detect if the AllowedValues string value starts with a negative number.

### DIFF
--- a/boto/rds/parametergroup.py
+++ b/boto/rds/parametergroup.py
@@ -146,7 +146,12 @@ class Parameter(object):
             value = int(value)
         if isinstance(value, int) or isinstance(value, long):
             if self.allowed_values:
-                min, max = self.allowed_values.split('-')
+                if self.allowed_values.startswith('-'):
+                    # range which begins with a negative value
+                    ign, min, max = self.allowed_values.split('-')
+                    min = '-' + min
+                else:
+                    min, max = self.allowed_values.split('-')
                 if value < int(min) or value > int(max):
                     raise ValueError('range is %s' % self.allowed_values)
             self._value = value


aheinz-fe:
Applied patch to homebrew-installed ansible/2.1.1.0 and tested on osx 10.11.6. 👍 
Also applied patch to homebrew-installed ansible19/1.9.6 and tested on osx 10.11.6. 👍

mcclurmc:
I'd really appreciate this patch being merged. This breaks my Ansible RDS param group playbooks and I have to apply this patch to my boto install, which I'd really prefer not to do.

GaboFDC:
Why is this PR still on hold? this is totally necessary can't automate the rds parameters groups, for postgres 9.6 without this.
Please some reply

mihu:
Ping. This should have been already merged. So sad.

GaboFDC:
No reply and none assignee?
Why is this PR so abandoned? what should we do for this to be taken into account?

PR #3546, opened 2016-05-24
Fix python3 compability

stBecker:
response.read() returns a byte-string
in python3. This breaks the assert on
line 326


listingmirror:
I have a similar PR at https://github.com/boto/boto/pull/3595

This way may be better.. one or the other needs merged though

jphaskins:
This is better then commenting out the assertion altogether, which is what I was doing before I saw this patch. Please accept.

eric-zhu-uw:
Would it be possible to get this PR merged? Currently, there's no alternative to `boto` to access MWS since there is no support in `boto3` and this makes it impossible to get reports with `python3` unless we fork.

PR #3550, opened 2016-05-30
Add ability to download MWS reports without storing them in memory

doihaveto:
Some MWS reports can weigh 400mb or more. The `get_report` method stores the report into a variable, potentially using up the system's memory. This fix allows handling the report using the socket object, so you can save the report to a file or handle it in chunks.
Usage example:

```
response = conn.get_report(ReportId=report_id, raw_response=True)
with open('report.txt', 'wb') as f:
    while True:
        chunk = response.read(1024 * 16)
        if not chunk:
            break
        f.write(chunk)

```

### DIFF
--- a/boto/mws/connection.py
+++ b/boto/mws/connection.py
@@ -297,7 +297,8 @@ class MWSConnection(AWSQueryConnection):
     def _required_auth_capability(self):
         return ['mws']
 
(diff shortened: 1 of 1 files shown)


disruptek:
Cute!  Thanks for the contribution!

disruptek:
Added to my branch; thanks.

PR #3553, opened 2016-06-06
Respect is_secure parameter in generate_url_sigv4

tipabu:
Previously, the protocol was hardcoded to https.


jamesls:
:shipit:  Looks good to me.

PR #3554, opened 2016-06-08
Add missing MWS methods

doihaveto:
Added some missing methods from the MWS fulfilment inbound shipment API

### DIFF
--- a/boto/mws/connection.py
+++ b/boto/mws/connection.py
@@ -611,6 +611,56 @@ class MWSConnection(AWSQueryConnection):
         """
         return self._post_request(request, kw, response)
 
+    @requires(['ShipmentId', 'IsPartnered', 'ShipmentType', 'TransportDetails'])
+    @boolean_arguments('IsPartnered')
+    @structured_objects('TransportDetails', members=True)
+    @api_action('Inbound', 30, 2)
+    def put_transport_content(self, request, response, **kw):
+        """Sends transportation information to Amazon about an inbound
+           shipment.
+        """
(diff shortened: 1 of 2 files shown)


disruptek:
Excellent.  Did you have any thorny problems with crafting the response parser?  I think I follow the change you made for ComplexDimensions...

Unfortunately, the maintainers will want some kind of test.  A test that produces an error message from MWS might be sufficient.  I haven't seen MWS patches accepted in quite awhile, though.

doihaveto:
The existing tests already cover it, as far as I can tell.

The response parser was pretty straightforward, just followed the MWS documentation and checked with a live connection to see that it parsed correctly.

disruptek:
I'm going to include this in my branch, with some changes as follows:

The `ComplexDimensions` type is used only in `ItemAttributes`, as you might see on an Amazon product page -- item dimensions and package dimensions.  We call it "complex" because each of the dimensions arrives with a units value stored as an XML attribute.  We map each dimension to what is essentially a hash with `Value` and `Units` keys.

In the case of this FBA call, the dimensions response element arrives with no XML attributes and the `Unit` element as a peer of any `Height`, `Length`, `Width` dimensions.  There is no `Weight` value sent, as it appears elsewhere in the response.

Upon reflection, the only reason to write a specific parser for this type is to cast the dimension values to decimals and to enable a pretty string representation.  The upshot of all this is that I'm going to parse this closer to the wire and call it `SimpleDimensions`.  The response will be easier to grok and easier to use.

PR #3562, opened 2016-06-22
Remove the broken link to PDF's

kyleknap:
Clicking on the "PDF Version" link currently results in 404: http://boto.readthedocs.io/en/latest/. I still believe that read the docs produces a PDF but when they changed the domain to readthedocs.io that broke that link. I much rather rely on the link to the PDF they provide in their widget instead of fix the link and have it break again.

cc @jamesls @JordonPhillips


jamesls:
:shipit:

PR #3564, opened 2016-06-27
Update endpoints.json

JordonPhillips:
cc @kyleknap @jamesls

### DIFF
--- a/boto/auth.py
+++ b/boto/auth.py
@@ -59,7 +59,9 @@ SIGV4_DETECT = [
     '.eu-central',
     '-eu-central',
     '.ap-northeast-2',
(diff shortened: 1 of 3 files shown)


kyleknap:
I think you need to update s3 to default to sigv4 as well.

JordonPhillips:
Gotcha

kyleknap:
It would also not hurt to add tests like this one: https://github.com/boto/boto/blob/develop/tests/unit/auth/test_sigv4.py#L553. Otherwise, looks good.

JordonPhillips:
Done

kyleknap:
:ship:

PR #3569, opened 2016-07-08
Mock https connection in ecs test

jamesls:
Makes this consistent with the EC2 unit tests.

cc @kyleknap @JordonPhillips

PR #3570, opened 2016-07-08
Fix ec2/ecs unit tests

jamesls:
Run ec2/ecs tests as part of default test run

I noticed that these tests aren't being run because they're
currently marked as executable, and nose skips executable files
by default.

After `-x` the files, I had an ec2 test failure because
`cmp` is now longer valid in python3.

I also noticed we're using `httplib` which fails in python3.

cc @kyleknap @JordonPhillips 

Note: I discovered this while working on #3569, so it pulls in that commit as well.


kyleknap:
:ship:

PR #3574, opened 2016-07-12
boto.dynamodb2.table.Table#batch_get() fails to paginate results if provisioned throughput is exceeded

collinstocks:
Fixes https://github.com/boto/boto/issues/3575.

From the [documentation](http://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_BatchGetItem.html) (emphasis mine):
> ### UnprocessedKeys
> A ***map of tables*** and their respective keys that were not processed with the current response. The UnprocessedKeys value is in the same form as RequestItems, so the value can be provided directly to a subsequent BatchGetItem operation. For more information, see RequestItems in the Request Parameters section.
>
> ***Each element*** consists of:
>
> - Keys - An array of primary key attribute values that define specific items in the table.
> - AttributesToGet - One or more attributes to be retrieved from the table or index. By default, all attributes are returned. If a requested attribute is not found, it does not appear in the result.
> - ConsistentRead - The consistency of a read operation. If set to true, then a strongly consistent read is used; otherwise, an eventually consistent read is used.
>
> If there are no unprocessed keys remaining, the response contains an empty UnprocessedKeys map.
>
> Type: String to KeysAndAttributes object map

### DIFF
--- a/boto/dynamodb2/table.py
+++ b/boto/dynamodb2/table.py
@@ -669,7 +669,7 @@ class Table(object):
         should be fetched)
 
         Returns an ``Item`` instance containing all the data for that record.
-        
+
         Raises an ``ItemNotFound`` exception if the item is not found.
 
         Example::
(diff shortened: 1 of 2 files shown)


collinstocks:
Could somebody please respond? `Table#batch_get()` is effectively broken in boto. This PR fixes it.

jamesls:
:shipit: looks good to me.

PR #3584, opened 2016-08-04
Amazon AWS S3 - Implement new Cross Region Replication feature

tpodowd:
- PUT/GET/DELETE Bucket replication
- Add test cases for Cross Region Replication feature


tpodowd:
Here's output from the new test case for CRR.

    [tpodowd@TMPro s3 (cross_region_replication)]$ nosetests-2.7 test_bucket:S3BucketCRRTest
    .
    Ran 1 test in 15.514s
    
    OK

PR #3589, opened 2016-08-13
add support target_groups for autoscaling. and inscale protection

blacked:
This PR add target groups support for autoscaling.
Also InScale protection for new instances.


adubkov:
I would like to have it merged. We have to maintain our own fork of boto just for target groups now...

Issue #3544, opened 2016-05-21
mws get_report python 3 regression


lmbrian:
Boto== 2.40
Python=3.5.1
OS = windows or ubuntu

calling `get_report` method for a simple `_GET_AFN_INVENTORY_DATA_` report

End up with a stacktrace like

```
Traceback (most recent call last):
  File "/webapps/my_package/lib/python3.5/site-packages/celery/app/trace.py", line 240, in trace_task
    R = retval = fun(*args, **kwargs)
  File "/webapps/my_package/lib/python3.5/site-packages/celery/app/trace.py", line 438, in __protected_call__
    return self.run(*args, **kwargs)
  File "/webapps/my_package/my_package/my_app/tasks.py", line 918, in update_fba_qty
    afn_inventory_result = amazon_api.get_report(integration_config, generate_report_id)
  File "/webapps/my_package/my_package/my_app/apis/amazon_api.py", line 109, in get_report
    MarketplaceId=US_MARKETPLACE.marketplace_id)
  File "/webapps/my_package/lib/python3.5/site-packages/boto/mws/connection.py", line 158, in requires
    return func(*args, **kw)
  File "/webapps/my_package/lib/python3.5/site-packages/boto/mws/connection.py", line 252, in wrapper
    return func(self, request, response, *args, **kw)
  File "/webapps/my_package/lib/python3.5/site-packages/boto/mws/connection.py", line 503, in get_report
    return self._post_request(request, kw, response)
  File "/webapps/my_package/lib/python3.5/site-packages/boto/mws/connection.py", line 326, in _post_request
    assert content_md5(body) == digest
AssertionError

```

Same everything works on Python 2.7 line.

listingmirror:
Totally just a equality of strings vs bytes dealio.


https://github.com/listingmirror/boto/commit/2b973e7ad0a1145d89ddf06621580753234af135

Happy to write up a test and patch, but not sure the "pythonic" way to do this in a join python2 / python3 codebase

listingmirror:
@canassa what is the upside to sometimes returning MD5 in bytes and sometimes in String? Seems like MD5 at heart is a String, we know the encoding of it...

canassa:
@listingmirror You mean by making sure that the `content_md5` always return a utf-8 string? It makes sense IMO. I will update my PR

listingmirror:
Hi,

Would would it take to get this merged? Do you guys want a PR or is there other plans for fixing this?

jphaskins:
[...]

PR #3595, opened 2016-08-24
Make MWS compatible with Python3

listingmirror:
This fixes https://github.com/boto/boto/issues/3544


listingmirror:
Another version of https://github.com/boto/boto/pull/3546

Issue #3598, opened 2016-09-02
boto.mturk.connection.get_all_hits(): TypeError: 'float' object cannot be interpreted as an integer


clarkgrubb:
I get a TypeError when I call `boto.mturk.connection.get_all_hits()`.  I am using Python 3.5 on Mac OS X El Capitan with boto 2.42.  My Mechanical Turk account has one HIT assignment pending review:

    $ cat Makefile 
    .DEFAULT_GOAL := run

    ve:
    	virtualenv -p python3 ve
    	. ve/bin/activate && pip install -r requirements.txt

    run: ve
    	. ve/bin/activate && python minimal.py

    $ cat requirements.txt 
    boto==2.42.0

    $ cat minimal.py 
    import os

    from boto.mturk.connection import MTurkConnection

    AWS_ACCESS_KEY_ID = os.environ['AWS_ACCESS_KEY_ID']
    AWS_SECRET_ACCESS_KEY = os.environ['AWS_SECRET_ACCESS_KEY']
    SANDBOX_HOST = 'mechanicalturk.sandbox.amazonaws.com'

    connection = MTurkConnection(aws_access_key_id=AWS_ACCESS_KEY_ID,
                                 aws_secret_access_key=AWS_SECRET_ACCESS_KEY,
                                 host=SANDBOX_HOST)

    print(connection.get_all_hits())

    $ make version
    uname -a
    Darwin Clarks-MacBook-Pro.local 15.6.0 Darwin Kernel Version 15.6.0: Thu Jun 23 18:25:34 PDT 2016; root:xnu-3248.60.10~1/RELEASE_X86_64 x86_64
    python --version
    Python 2.7.12

    $ make
    virtualenv -p python3 ve
    Running virtualenv with interpreter /usr/local/bin/python3
    Using base prefix '/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5'
    New python executable in ve/bin/python3.5
    Also creating executable in ve/bin/python
    Installing setuptools, pip...done.
    . ve/bin/activate && pip install -r requirements.txt
    You are using pip version 6.1.1, however version 8.1.2 is available.
    You should consider upgrading via the 'pip install --upgrade pip' command.
    Collecting boto==2.42.0 (from -r requirements.txt (line 1))
      Using cached boto-2.42.0-py2.py3-none-any.whl
    Installing collected packages: boto
    Successfully installed boto-2.42.0
    . ve/bin/activate && python minimal.py
[...]

PR #3599, opened 2016-09-02
Mturk get_pages() fix

clarkgrubb:
This seems to fix the probem in https://github.com/boto/boto/issues/3598 for me.


awatts:
Doesn't changing `itertools.imap` to `map` break Python 2 in favor of working on Python 3 then? Shouldn't you add

```python
from boto.compat import six
```

in the imports, and then change the instances of `itertools.imap` to `six.moves.map`? That seems to be the boto style for handling compatibility.

PR #3603, opened 2016-09-09
Add MWSAuthToken support

cmeyer90:
Simply push the MWSAuthToken as a keyword argument when creating an mws connection.

Ex:

mws = MWSConnection(mAccessKey, mSecretKey, MWSAuthToken='token')


disruptek:
Added to my branch; thanks.

PR #3607, opened 2016-09-15
Fixed GS connection converting bucket_name

Rsgm:
Boto was incorrectly converting bucket_name to bytes. It should be kept
a string. This caused an error down the stacktrace in boto.s3.connection:94 when
trying to build the path.


Rsgm:
I am closing this for now, since there are a few other errors of the same type that I missed.

PR #3628, opened 2016-10-17
Fix asadmin and https proxy issues

bandesz:


bandesz:
sorry, wrong repository

PR #3649, opened 2016-12-08
Update endpoints

JordonPhillips:
Update endpoints, including support for ca-central-1

cc @kyleknap @stealthycoin


boto/endpoints.json
```diff
@@ -451,9 +495,10 @@
     "s3": {
         "ap-northeast-1": "s3-ap-northeast-1.amazonaws.com",
         "ap-northeast-2": "s3.ap-northeast-2.amazonaws.com",
-        "ap-south-1": "s3-ap-south-1.amazonaws.com",
```
kyleknap: Looks like the endpoint using the dot instead of the dash is valid, so should be fine that this changes.

PR #3650, opened 2016-12-12
Added support for Region-based Reserved Instances

pmp:
When using default api_version parameters during the creation of an EC2Connection object, the API will not return the new Region-based Reserved Instances, but if you invoke with api_version='2016-11-15', then the internals of boto will have access to this information in the raw XML feed from the AWS apis. This change will simply process the 'scope' tags and include that in the ReservedInstancesOffering object.


wenhongqiang:
@pmp  anyone can check this? or merge this pr? thanks.

pmp:
I would love for this to be merged in, but I am not in control of this...

robinske:
Also adding my support for this!

PR #3655, opened 2016-12-15
Allow boto2 to delete a "superbatch" of messages

MPagel:
What?
====
delete_message_batch_no_object and helper functions added

What this allows you to do
======
Lifts the Boto requirement of only deleting messages that are still encased in Message objects.
Also allows you to delete more than 10 messages in a superbatch

What was the issue with the old code that this patch addresses
========
delete_message_batch expects a Message class object
Copying the Boto source for delete_message_batch and allowing it to use a non-Message object (ala boto3) also fails if you're deleting more than 10 "messages" at a time.

Workflow
=======
Break messages in an iterable object (containing at least an 'id' and 'receipt_handle') into batches of 10
Issue delete command for each group of up to 10
Discover how many objects were successfully deleted or failed to delete of the 10 and compile a CSV-style list of which ones fell into each category.
Return the results in a string, simplified if all were successfully deleted or all failed.

### DIFF
--- a/boto/sqs/connection.py
+++ b/boto/sqs/connection.py
@@ -28,6 +28,8 @@ from boto.sqs.attributes import Attributes
 from boto.sqs.batchresults import BatchResults
 from boto.exception import SQSError, BotoServerError
 
+from itertools import islice
+import cStringIO
 
 class SQSConnection(AWSQueryConnection):
     """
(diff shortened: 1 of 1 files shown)


MPagel:
changed target of pull request to `develop` from `master`

PR #3657, opened 2016-12-19
Migrate endpoints

JordonPhillips:
This PR migrates boto to the new endpoints format used by botocore/boto3. Note that it does NOT add support for endpoint heuristics. This PR is already massive, so I'd rather pull out as much into other prs as possible. All it does change the underlying data format used while maintaining the exact behavior as before. Forthcoming PRs will add the ability to connect to regions not explicitly defined as well as update the related scripts.

For the sake of readability, I've broken this out into several commits. I would recommend going through one by one.


boto/exception.py
```diff
@@ -583,3 +583,18 @@ class InvalidInstanceMetadataError(Exception):
     def __init__(self, msg):
         final_msg = msg + '\n' + self.MSG
         super(InvalidInstanceMetadataError, self).__init__(final_msg)
+
+
+class BaseEndpointResolverError(Exception):
```
JordonPhillips: Since this is "shared" code I would not be in favor of removing them in case we have to backport fixes. I could see moving them in with the `EndpointResolver` itself, is that what you had in mind?

kyleknap: Yeah I was thinking if we can consolidate these along with ``EndpointResolver`` in the same location so it is really easy to do backports. Having them in ``exceptions`` makes that slightly harder.

tests/unit/test_regioninfo.py
```diff
@@ -46,11 +46,7 @@ def setUp(self):
 
     def test_load_endpoint_json(self):
         endpoints = load_endpoint_json(boto.ENDPOINTS_PATH)
-        self.assertTrue('ec2' in endpoints)
-        self.assertEqual(
-            endpoints['ec2']['us-east-1'],
-            'ec2.us-east-1.amazonaws.com'
-        )
+        self.assertTrue('partitions' in endpoints)
```
JordonPhillips: What do you mean? This is changing the entire format.

boto/endpoints.py
```diff
@@ -0,0 +1,415 @@
...
+import re
+
+from boto.exception import NoRegionError
+
+LOG = logging.getLogger(__name__)
+DEFAULT_URI_TEMPLATE = '{service}.{region}.{dnsSuffix}'
+DEFAULT_SERVICE_DATA = {'endpoints': {}}
+
+
+class BaseEndpointResolver(object):
```
JordonPhillips: That makes sense

tests/unit/test_regioninfo.py
kyleknap: Nevermind ``load_endpoints_json`` is misleading. Did not realize that all it interface does is load a json file. There is nothing specific to endpoints in that function. The change should be fine.

boto/endpoints.py
```diff
@@ -0,0 +1,415 @@
...
+                result[key] = from_data[key]
+
+    def _expand_template(self, partition, template, service_name,
+                         endpoint_name):
+        return template.format(
+            service=service_name, region=endpoint_name,
+            dnsSuffix=partition['dnsSuffix'])
+
+
+class BotoEndpointResolver(EndpointResolver):
```
JordonPhillips: I don't think there's much that can be made private without combining the concerns of the static endpoint builder and the endpoint resolver. There are four additional methods exposed here:

* `is_global_service`: this can be removed with no issue
* `resolve_hostname`: this can also be removed since it's a convenience method, but then we would have to perform its logic in at least two places (the endpoint builder and the region connection code). I would rather keep it, but it could be removed.
* `get_available_services`: to generate the static endpoint list we need to know what all the services are. This could be replaced with a static list.
* `get_all_available_regions`: I don't think this can be removed. We need to know what endpoints are valid to build the static list. I also use that to enforce no regions are guessed, though I could potentially do that in other ways.

I could make a `StaticEndpointResolver` that subclasses `BotoEndpointResolver` which does not use the endpoint regex and which also includes the `build` function, but it seems like sharing unnecessary concerns.

boto/endpoints.py
```diff
@@ -0,0 +1,415 @@
...
+                regions.update(partition['regions'].keys())
+            else:
+                regions.update(
+                    super(BotoEndpointResolver, self).get_available_endpoints(
+                        endpoint_prefix, partition_name
+                    )
+                )
+
+        # boto2 endpoint format has no partitions, so load them all.
+        boto_endpoints = self._get_available_boto_endpoints(service_name)
```
JordonPhillips: I'm not sure what you mean. I can't simply merge the files since they have very different formats. I don't have enough information to port the old format to the new, and I would lose too much information porting the new to the old.

kyleknap: Hmm that makes sense.

I wonder if that begs the question: should the override be in the endpoint resolver logic? Instead should it be an override in the static endpoint builder? I think it would make sense from an interface perspective of the static builder if the interface was:
* It will use the default heuristics to generate a static list
* Then a user can provide an additional set of data to merge over whatever the default heuristics generate

It would make the code much cleaner as you would also be dealing with the same data format.

Additionally, I am not sure if the ``BotoEndpointResolver` necessarily need to have that override logic. It is just going to be a fall back if nothing matches from the statically built list when connection to a region. Right?

boto/endpoints.py
kyleknap: Yeah, I think keeping ``get_all_available_regions``, ``get_all_available_regions``, and then some general ``get/resolve_endpoint`` method (``resolve_hostname`` might fit this bill as well) that is used as the fallback for the static list would be fine.

I was mainly concerned about the ones that leak implementation details like ``is_global_service()`` that are not too useful from an external user perspective who really only care about what endpoint they should be using based on their region.

boto/endpoints.py
JordonPhillips: No, `BotoEndpointResolver` will not be a fallback, it will be used directly. It's pretty silly to generate the endpoint for every service in every region when you only want a single service in a single region. There's no caching like we have in botocore, so we would need to do that *every* time a client is created.

My [current strategy](https://github.com/JordonPhillips/boto/blob/heuristic-endpoints/boto/regioninfo.py#L150) is to resolve using the boto resolver and then to check if the region exists for the service if necessary. I could also override [this](https://github.com/boto/boto/pull/3657/files#diff-c0b2964c406eac309891c4e25c7ded86R155) and have its functionality be dependent on an argument.

I think the overrides should live in the boto resolver. It's the job of that class to resolve endpoints in the way boto expects, it's not the job of the builder.

boto/endpoints.py
JordonPhillips: I went ahead and made `is_global_service` private. I had left it public to have tests for it since I added it before it was used. The particular edge cases are now covered by other tests, however.

JordonPhillips:
Updated. Botocore endpoint resolver is vendored and `is_global_service` was made private.

boto/vendored/regions/regions.py
```diff
@@ -0,0 +1,188 @@
+# Copyright 2014 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
JordonPhillips: The idea was to keep it as close to the source as possible to make backporting bug fixes simpler.

boto/vendored/regions/regions.py
```diff
@@ -0,0 +1,188 @@
...
+        if region_name in service_data['endpoints']:
+            return self._resolve(
+                partition, service_name, service_data, region_name)
+        # Check to see if the endpoint provided is valid for the partition.
+        if self._region_match(partition, region_name):
+            # Use the partition endpoint if set and not regionalized.
+            partition_endpoint = service_data.get('partitionEndpoint')
+            is_regionalized = service_data.get('isRegionalized', True)
+            if partition_endpoint and not is_regionalized:
+                LOG.debug('Using partition endpoint for %s, %s: %s',
```
JordonPhillips: I'm not a fan of having this be any different than what's in botocore. We should be able to just turn them off.

boto/endpoints.py
JordonPhillips: I don't think it makes sense to leak the concerns of resolving an endpoint to a builder. I could understand moving it to a different class, but then you'll have to re-implement the prioritization logic everywhere or create a third class that stitches the two together.

We won't be able to side-step the fact that we have to support both formats side-by-side. Currently the legacy format is only touched directly by functions that deal only with that format. The names of said functions specifically call out that they're working with that format. I agree with your other comment that it should be referenced in a more consistent way, and think that will help any confusion. The new format is only touched directly in cases where needed information is not provided by any existing functions. We could move that functionality up into the `EndpointResolver` to eliminate touching it directly at all.

Regarding the fallback question, I don't think the risk is really impacted either way. We still have to go through all of the new code to resolve all the regions. In fact, we have to go through *more* of it to resolve everything (since we need the static endpoint builder in addition to the endpoint resolver).

Furthermore, resolving every endpoint for every service is incredibly slow. For a known region it's ~4.5x slower than making a single call to `resolve_hostname` and for an unknown region it's a full order of magnitude slower.

jamesls: > I don't think it makes sense to leak the concerns of resolving an endpoint to a builder. I could understand moving it to a different class, but then you'll have to re-implement the prioritization logic everywhere or create a third class that stitches the two together

I want to make sure we're on the same page in terms of what's being proposed.  I think the main disconnect here is I (and I believe Kyle as well) are suggesting to *not* expose a "BotoEndpointResolver" as a first class concept.  I propose that the class should *should not even* exist.  

Why?

Consider `regioninfo.load_regions()`.  The only thing that matters to callers of that function is that we return a dictionary of `{service: {region: endpoint}}`.  How we construct that is an implementation detail, as long as we return that dict we can create it however we want.

So what's the least invasive/simplest way to make this change?

Right now the code in the `develop` branch is this:

```python
def load_regions():
    endpoints = load_endpoint_json(boto.ENDPOINTS_PATH)
    additional_path = None
    if os.environ.get('BOTO_ENDPOINTS'):
        additional_path = os.environ['BOTO_ENDPOINTS']
    elif boto.config.get('Boto', 'endpoints_path'):
        additional_path = boto.config.get('Boto', 'endpoints_path')
    if additional_path:
        additional = load_endpoint_json(additional_path)
        endpoints = merge_endpoints(endpoints, additional)
    return endpoints
```

The first line of that function loads a JSON file that's already in the exact structure we need.  So the simplest way to change this would be to change that first line to:

```
def load_regions():
    endpoints = construct_legacy_format_using_botocore_endpoints_format()
    ... everything else is the same ...
```

Where the job of this new function is to take botocore's endpoints.json and transform them into the legacy format.  It has a single job.  It does know about `BOTO_ENDPOINTS` or overrides or anything.  Take botocore endpoints -> produce legacy endpoint format.  That's it.

Let's say we stopped there and forgot about doing endpoint heuristics.

This is a pretty low risk change.  In fact we could pretty definitively know that nothing will break.  Ensure that `load_regions()` in the develop branch and `load_regions()` in this PR return the exact same results.  If that's the case we're fine.

It's also really simple to review.  With the exception of one line of code changed in `load_regions()` everything else is entirely brand new code that's not touching anything else in boto.  And the new code is purely about transforming one format to another.

JordonPhillips: I did more performance testing. For the following numbers I'm constructing an `awslambda` client in `us-east-1` and using the current state of `develop` as a baseline. Delegating all resolution to `BotoEndpointResolver` results in a 1.3x slowdown. Building out endpoints every time results in a 5x slowdown. This is on the order of milliseconds.

If we think that's an acceptable performance hit, we can move on. I've updated to move to that strategy.

JordonPhillips:
Updated and ready to be looked at again.

boto/vendored/regions/exceptions.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
kyleknap: So the one thing that I am afraid of is if we update the vendored code and we forget to add these exceptions back, it could lead to runtime errors that are harder to find.

I really like this approach mainly because it is the easiest and straightforward to maintain as it requires the least amount of manual steps. Specifically there is no build step. All we have to do is copy and paste the file over. If we try to combine these two, there will be steps requiring to copy the exception into the regions.py file and remove the imports. So if we ever update this file again, I can see it be really easy to forget that step and I would not want this PR merged until there are tests in place to make sure we did the revendoring correctly, such as checking the exceptions.  Another bonus of keeping it this way is that if we do the copy and paste of the file over and we forget an exception or some other import, we will get import exceptions that will definitely cause the test to fail, without us needing specific tests for the vendored code.

boto/vendored/regions/regions.py
kyleknap: I actually originally thought of it just being in ``vendored`` as well, but the additional ``regions`` module is growing on me. I like the fact that if I am trying to update the vendored code related to regions, I know exactly where to look, especially if we have the exceptions and possibly have more modules as revendoring of this logic occurs.

@jamesls Is there any other concerns about the extra layer of nesting other than it looking necessary?

boto/endpoints.py
```diff
@@ -0,0 +1,200 @@
...
+            else:
+                regions.update(
+                    super(BotoEndpointResolver, self).get_available_endpoints(
+                        endpoint_prefix, partition_name
+                    )
+                )
+
+        return list(regions)
+
+    def construct_endpoint(self, service_name, region_name=None):
```
JordonPhillips: `resolve_hostname` pulls out the hostname / ssl common name (if present). We can't change the return type of `construct_endpoint` as it is an inherited method.

boto/endpoints.py
```diff
@@ -0,0 +1,200 @@
...
+        # A mapping can be provided that handles the mapping between
+        # "service names" and endpoint prefixes.
+        'awslambda': 'lambda',
+        'cloudwatch': 'monitoring',
+        'ses': 'email',
+        'ec2containerservice': 'ecs',
+        'configservice': 'config',
+    }
+
+    def __init__(self, endpoint_data, service_rename_map=None):
```
JordonPhillips: No. Actually, I should probably make that default private. We need both the boto2 -> endpoint_prefix mapping and the endpoint_prefix -> boto2 mapping. To avoid reversing the mapping every time it's necessary, it is reversed when the class is constructed.

boto/endpoints.py
```diff
@@ -0,0 +1,200 @@
...
+        # Given a service name, 'ec2', build a dict of
+        # 'region' -> 'hostname'
+        endpoints = {}
+        regions = self._resolver.get_all_available_regions(service_name)
+        for region_name in regions:
+            endpoints[region_name] = self._resolver.resolve_hostname(
+                service_name, region_name)
+        return endpoints
+
+    def _handle_special_cases(self, static_endpoints):
```
JordonPhillips: The cloudsearchdomain endpoints are irrelevant, they aren't used. A custom hostname must always be supplied. See [here](https://github.com/boto/boto/blob/develop/boto/cloudsearchdomain/layer1.py#L62) where CSD requires host vs [here](https://github.com/boto/boto/blob/develop/boto/awslambda/layer1.py#L62) where lambda takes from the region. The region endpoint used to be a fallback, but [this commit](https://github.com/boto/boto/commit/b7f8369d964774c3a2fdd3f4f18beff3b0762bc2) removed that and raised an error instead.

The endpoints are kept here to preserve existing functionality as close as possible.

JordonPhillips:
This should be good for further look / discussion

boto/endpoints.py
kyleknap: Ok that works. I just thought it was weird that there were two different interfaces for this.

boto/endpoints.py
kyleknap: Make sense.

boto/endpoints.py
kyleknap: Thinking about this again. I think it if fine keeping it here. The customization is specifically needed for that static endpoint file. The data for cloudsearchdomain does not exist in the endpoints file so it would be strange to inject it there.

boto/vendored/regions/exceptions.py
jamesls: I don't follow how this is any simpler.  You have to curate exceptions.py to pull in the correct exceptions do you not?  You also have to do this transitively to pull in all the base class exceptions as well.  If this just a literal `cp` of a file then sure, but you already have to do manual work, I'd prefer to do the simpler option of keeping everything in a single file.

boto/vendored/regions/regions.py
jamesls: Unnecessary complication.

We literally need one class here.

I think there's too much emphasis placed on ease of backporting here.  After the initial PR was merged in feb last year, we've literally made a single line change to this module in all of 2016.  Sure we may need to backport things, but at the frequency of once a year, I don't think it's worth all this complication for such an infrequent task.  I'd rather go with the simpler approach.

boto/endpoints.py
jamesls: Ah, I was about to ask about this.  In that case, why is this a subclass of `BotoEndpointResolver` if we're adding additional methods that (presumably) are going to be the preferred methods to use with this class (i.e `resolve_hostname` instead of `construct_endpoint`).  Seems simpler to just use delegation and only expose the methods that are needed.

JordonPhillips:
@jamesls The other two exposed methods (`get_available_endpoints` and `construct_endpoint`) are inherited methods.

boto/vendored/regions/exceptions.py
jamesls: Chatted offline about this.  Given how infrequently we've updated this class in botocore, I'm good with leaving the structure as is.

JordonPhillips:
@jamesls I looked into making `BotoEndpointResolver` just take a normal `EndpointResolver` in and use it. It doesn't seem like it will be possible without digging into its internals, for the following reasons:

* I need to know if a service is global within a partition. That can be hacked by searching for the string "global" in a service's available regions, but it's not ideal and there is no guarantee that will work forever.
* I need to know all the regions in a partition. The only way around that without cracking open the endpoints file manually is to get all the regions for all the services in the partition to accumulate that list. That's also really hacky, and makes the performance problems significantly worse.

I really don't think the tradeoff is worth it, especially since the `BotoEndpointResolver` **is** an `EndpointResolver`.

jamesls:
I looked into the code and I still don't think your points above negate the original feedback.  I don't think we should be exposing more than we need to in terms of a public API. 

> ...especially since the BotoEndpointResolver is an EndpointResolver.

It's unconventional to subclass something and create new public methods that consuming code is expected to use (and also not use the inherited public methods).  That is, consuming code *requires* type `BotoEndpointResolver`, and won't work with type `EndpointResolver`.  In which case, just create an interface that only has methods needed by consuming code.

JordonPhillips:
A potential solution is to just add the necessary interface (listing all the regions in a partition) to `EndpointResolver` itself. I think I would prefer that to mucking about with the endpoints file itself.

kyleknap:
Alright. After taking a break and looking at this a second time, what are everyone's thoughts on this:

1) Import the vendored ``EndpointResolver`` as ``_BotocoreEndpointResolver``. That also makes people shy away from using that directly.
2) Break ``BotoEndpointResolver`` into two classes:
   i) A ``_CompatBotocoreEndpointResolver`` class which is just a subclass of ``_BotocoreEndpointResolver`` with no extra methods like ``resolve_hostname``. We also make this is as internal as possible; we put the underscore in front of the class name and have wording in the docstring to not use this and it is for internal use only. It's sole purpose is to add compatibility logic on top of the ``_BotocoreEndpointResolver``
   ii) A ``EndpointResolver`` class which is public. It does not subclass from ``EndpointResolver`` and just has the methods ``resolve_hostname``, ``get_all_available_regions``, ``get_all_available_services``. The class would require no arguments to instantiate it. All of its arguments would just proxy to an instantiated ``_CompatBotocoreEndpointResolver`` or ``_BotocoreEndpointResolver``. This should just be a small wrapper that hides away some of the other interfaces from the ``BotocoreEndpointResolver`` and adds that ``resolve_hostname``

2) Then use this ``EndpointResolver`` for the static endpoint builder

Let me know what you think about this. I think it would not be difficult to add and it would firm up the public interfaces.

boto/endpoints.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+class BotoEndpointResolver(object):
+    """Resolves endpoint hostnames for AWS services.
+
+    This is NOT intended for external use.
+    """
+
+    def __init__(self, endpoint_data, service_rename_map=None):
+        self._resolver = _CompatEndpointResolver(
```
kyleknap: Probably would want to put the parameter docstrings in the ``__init__``.

tests/unit/test_endpoints.py
```diff
@@ -154,21 +145,6 @@ def test_get_all_regions_with_renames(self):
             'us-bar', 'eu-baz', 'us-foo', 'foo-1', 'foo-2', 'foo-3'])
         self.assertEqual(regions, expected_regions)
 
-    def test_construct_endpoint_with_rename(self):
```
kyleknap: Cool we have tests for renaming. Fine to remove then.

boto/endpoints.py
```diff
@@ -134,6 +130,53 @@ def _service_name(self, endpoint_prefix):
         return self._service_name_map.get(endpoint_prefix, endpoint_prefix)
 
 
+class BotoEndpointResolver(object):
+    """Resolves endpoint hostnames for AWS services.
+
+    This is NOT intended for external use.
```
kyleknap: I think this is appropriate wording. While you can use it, it is not recommended as it is not intended.

boto/endpoints.py
JordonPhillips: Added

kyleknap:
Cool. Should be good to merge then once the tests pass. Once you merge it. Also, make sure you rebase the rest of your PR's so it is a bit easier to review.

PR #3658, opened 2016-12-19
Update rebuild endpoints to new format

JordonPhillips:
This will now simply filter out the services we don't use in order to save space.

This uses a static list of supported services, but we could also use the current file as the source of truth. The advantage of having a static list is that it is protected from the case where the endpoints file gets into a bad state.

This depends on #3657. Only the last commit should be looked at in this PR.

### DIFF
--- a/scripts/rebuild-endpoints.py
+++ b/scripts/rebuild-endpoints.py
@@ -1,21 +1,8 @@
 """Rebuild endpoint config.
 
-Final format looks like this::
-{
-    "autoscaling": {
-        "ap-northeast-1": "autoscaling.ap-northeast-1.amazonaws.com",
-        "ap-northeast-2": "autoscaling.ap-northeast-2.amazonaws.com",
-        "ap-southeast-1": "autoscaling.ap-southeast-1.amazonaws.com",
-        ...
-    },
-    "service-name": {
-        "region": "hostname"
-    }
-}
-
-This will use the EndpointResolver from botocore to regenerate
-endpoints.  To regen the latest static endpoints, ensure you have
-the latest version of botocore installed before running this script.
+This will pull in the latest endpoints, filtering out services that aren't
+supported by boto. By default, this will pull in the endpoints from botocore,
+but a specific file can also be specified.
 
 Usage
 =====
@@ -23,6 +10,10 @@ Usage
 To print the newly gen'd endpoints to stdout::
 
     python rebuild-endpoints.py
+
+To print the newly gen'd endpoints in the legacy format::
+
+    python rebuild-endpoints.py --legacy-format
 
 To overwrite the existing endpoints.json file in boto:
 
@@ -39,14 +30,14 @@ import os
 import json
 import argparse
 
-
+# Use third party ordereddict for older versions of Python
 try:
-    import botocore.session
-    from botocore.regions import EndpointResolver
+    from collections import OrderedDict
 except ImportError:
-    print("Couldn't import botocore, make sure it's installed in order "
-          "to regen endpoint data.")
-    sys.exit(1)
+    from ordereddict import OrderedDict
+
+from boto.endpoints import BotoEndpointResolver
+from boto.endpoints import StaticEndpointBuilder
 
 
 EXISTING_ENDPOINTS_FILE = os.path.join(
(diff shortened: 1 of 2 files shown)


scripts/rebuild-endpoints.py
```diff
@@ -39,240 +30,83 @@
...
+    endpoints_data = load_endpoints(args.endpoints_file)
+    filter_services(endpoints_data)
+
+    if args.legacy_format:
+        builder = StaticEndpointBuilder(BotoEndpointResolver(endpoints_data))
+        endpoints_data = builder.build_static_endpoints(SUPPORTED_SERVICES)
+
     json_data = json.dumps(
-        static_endpoints, indent=4, sort_keys=True, separators=(',', ': '))
+        endpoints_data, indent=2, sort_keys=True, separators=(',', ': '))
```
kyleknap: I wonder if we should be sorting these or be retaining the original order of the endpoints file through an OrderedDict as the order in botocore will change in moving over to boto using ``sort_keys``. Only bringing it up because it would be nice for when we need to compare the endpoints.json in boto and botocore to have them as close as possible.

scripts/rebuild-endpoints.py
```diff
@@ -39,240 +30,83 @@
...
-                endpoints[region_name] = self._resolver.get_hostname(
-                    service_name, region_name)
-        return endpoints
-
-    def _deal_with_special_cases(self, static_endpoints):
-        # I'm not sure why we do this, but cloudsearchdomain endpoints
-        # use the exact same set of endpoints as cloudsearch.
-        if 'cloudsearch' in static_endpoints:
-            static_endpoints['cloudsearchdomain'] = static_endpoints['cloudsearch']
+def filter_services(endpoints):
```
kyleknap: Curious here and just throwing it out there. What would happen if we did not filter the services and just used the exact botocore file? It may be a little strange to have unsupported services in the file but may be less error prone if we can use a straight copy and paste of the file from botocore. What are your thoughts?

JordonPhillips: It increases the search time and the time it takes to generate the static endpoint map. Other than that I don't think it would actually cause any errors.

scripts/rebuild-endpoints.py
JordonPhillips: I'm pretty sure those are sorted, but it's still a good point. Updated.

scripts/rebuild-endpoints.py
kyleknap: That's a good point about speed in terms of searching/loading. We have tests in place so that we will not accidentally filter out a service (i.e. ensure that we have every single service still that existed in the old format). Correct?

I just want to be careful that we do not introduce regressions say we pull in an update to the endpoints file and introduce a regression due to an issue with our filtering.

boto/compat.py
```diff
@@ -29,6 +29,13 @@
     import json
 
 
+# Use third party ordereddict for older versions of Python
+try:
+    from collections import OrderedDict
+except ImportError:
+    from ordereddict import OrderedDict
```
kyleknap: Not sure if we want to have this logic here. From the looks of it ``boto ``does not have a dependency in ``setup.py`` for ``ordereddict`` so it may cause import errors... I would recommend just moving this to the script.

scripts/rebuild-endpoints.py
JordonPhillips: That's a fair point, I added a test to make sure no services are lost. It would be very difficult to verify no endpoints are lost, but as long as we have the heuristics in place we should be good there.

JordonPhillips:
Good to be looked at again.

JordonPhillips:
Rebased

scripts/rebuild-endpoints.py
```diff
@@ -39,240 +30,89 @@
...
-    strict_resolver = StrictEndpointResolver(resolver, endpoint_data)
-    builder = StaticEndpointBuilder(strict_resolver)
-    static_endpoints = builder.build_static_endpoints(
-        known_services_in_existing_endpoints)
+    endpoints_data = load_endpoints(args.endpoints_file)
+    filter_services(endpoints_data)
+
+    if args.legacy_format:
+        builder = StaticEndpointBuilder(BotoEndpointResolver(endpoints_data))
+        endpoints_data = builder.build_static_endpoints(SUPPORTED_SERVICES)
```
kyleknap: I think we want to remove the ``SUPPORTED_SERVICES`` as we already filter it before: https://github.com/boto/boto/blob/endpoints-v2/boto/endpoints.py#L199 and the docstring says it needs to be boto2 naming scheme.

scripts/rebuild-endpoints.py
```diff
@@ -39,240 +30,89 @@
...
-        # This can be accomplished by using this method and
-        # regions_for_partition. See the _special_case_global_service
-        # method in StaticEndpointBuilder.
-        for partition_name in self.get_available_partitions():
-            if self.is_service_in_partition(service_name, partition_name):
-                yield partition_name
-
-    def get_available_partitions(self):
-        return self._resolver.get_available_partitions()
+SUPPORTED_SERVICES = [
```
kyleknap: Might be good to point out that this is botocore naming scheme not boto2 naming scheme. That tripped me up a bit.

tests/unit/test_endpoints.py
```diff
@@ -294,3 +296,33 @@ def __init__(self, service, old_endpoints, new_endpoints):
...
+        'sdb', 'ses', 'sns', 'sqs', 'storagegateway', 'sts', 'support', 'swf'
+    ]
+
+    with open(boto.ENDPOINTS_PATH) as f:
+        endpoints = json.load(f)
+
+    resolver = BotoEndpointResolver(endpoints)
+    available_services = resolver.get_available_services()
+
+    for service in supported_services:
```
kyleknap: We talked about expanding this to include regions and hostnames so the current endpoint in the data directory will not regress for users that did not opt in to it.

tests/unit/test_endpoints.py
```diff
@@ -294,3 +295,26 @@ def __init__(self, service, old_endpoints, new_endpoints):
...
+
+    builder = StaticEndpointBuilder(BotoEndpointResolver(new_endpoints))
+    built = builder.build_static_endpoints()
+
+    # Assert no services are lost
+    for service, service_endpoints in old_endpoints.items():
+        new_service_endpoints = built.get(service, {})
+        for region, regional_endpoint in service_endpoints.items():
+            new_regional_endpoint = new_service_endpoints.get(region)
+            yield assert_equal, regional_endpoint, new_regional_endpoint
```
kyleknap: Might be good to have a more explicit error message that mentions what service and region was being used on fail for this assertion.

kyleknap:
Looks good. Go ahead and merge it once the build passes.

PR #3664, opened 2017-01-05
Unify connection and allow using endpoint heuristics

JordonPhillips:
This is dependent on #3657, that should be looked at first.

This creates a generic connect function to point all the other connect functions at. This generic connect function supports using endpoint heuristics if necessary.

### DIFF
--- a/boto/awslambda/__init__.py
+++ b/boto/awslambda/__init__.py
@@ -20,6 +20,7 @@
 # IN THE SOFTWARE.
 #
 from boto.regioninfo import RegionInfo, get_regions
+from boto.regioninfo import connect
 
 
 def regions():
@@ -34,7 +35,6 @@ def regions():
 
 
 def connect_to_region(region_name, **kw_params):
-    for region in regions():
-        if region.name == region_name:
-            return region.connect(**kw_params)
-    return None
+    from boto.awslambda.layer1 import AWSLambdaConnection
+    return connect('awslambda', region_name,
+                   connection_cls=AWSLambdaConnection, **kw_params)
--- a/boto/s3/__init__.py
+++ b/boto/s3/__init__.py
@@ -24,6 +24,7 @@
 #
 
 from boto.regioninfo import RegionInfo, get_regions
+from boto.regioninfo import connect
 
 
 class S3RegionInfo(RegionInfo):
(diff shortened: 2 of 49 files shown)


JordonPhillips:
Rebased you should be looking at the last 3 commits

boto/awslambda/__init__.py
```diff
@@ -28,13 +30,10 @@ def regions():
     :rtype: list
     :return: A list of :class:`boto.regioninfo.RegionInfo`
     """
-    from boto.awslambda.layer1 import AWSLambdaConnection
```
kyleknap: Was there a reason that we imported the connection object within the ``regions()`` function? Maybe @jamesls would know? Seems odd that this would be the pattern chosen without a particular reason.

jamesls: I don't remember why.  I was going to guess to avoid importing all of boto including connection classes on an `import boto`, but the various `boto.connect_<service>`_ functions in `boto/__init__.py` import the connection classes inline, so these are probably unnecessary.

However, my previous stance here still remains for boto which is to be conservative in the changes unless there's a really good reason not to do so vs. making changes that seem harmless but not actually knowing for sure.

JordonPhillips:
Updated

JordonPhillips:
Rebased

boto/s3/__init__.py
```diff
@@ -59,16 +60,17 @@ def regions():
...
-                return region.connect(**kw_params)
-            # If it is nothing then remove it from kw_params and proceed with default
-            else:
-                del kw_params['host']
-        if region.name == region_name:
+    from boto.s3.connection import S3Connection
+    if 'host' in kw_params:
+        if kw_params['host'] not in ['', None]:
+            region = S3RegionInfo(
+                name='custom',
```
JordonPhillips: It was arbitrary. The old logic would just take whatever happened to be the first returned connection and modify it. So the region name could be 'us-east-2' or 'eu-central-1' depending on how the regions were ordered. In any case the value isn't useful here.

kyleknap: Right. That's quirky. My best guess it would currently be ``ap-northeast-1`` region if the list stays ordered. What would happen if we set ``custom`` to the region they provided? Will that mess up any s3 connection code? That seems like the most appropriate way to handle it, especially if sigv4 signing is involved where region matters.

JordonPhillips:
I think introspecting each module to try to find a connect function is a little nuts. I simply found all usages of `get_regions` and modified where necessary.I can add tests for each one of those, however.

kyleknap:
Might have worded that poorly. So I don't think we need to introspect the modules. We still would have to list the modules out manually. I suggested just doing the one test file with that list to make it a little less tedious, but having a test for each module/connect_to_region function is good as well. I think once we have that in place. We will be more certain that nobody gets broken by changing the ``connect_to_region`` function for each service.

boto/s3/__init__.py
JordonPhillips: Yeah, it doesn't even pass the region along. [S3 region connect](https://github.com/boto/boto/blob/endpoints-v2/boto/s3/__init__.py#L43) vs [standard region connect](https://github.com/boto/boto/blob/endpoints-v2/boto/regioninfo.py#L223).

JordonPhillips:
Added tests. Found a few bugs!

JordonPhillips:
Squashed some commits, will merge once tests pass.

PR #3668, opened 2017-01-18
Corrected dynamodb2 query_count()

leighsmith:
query_count() could exit early returning an incorrect count if all entries were filtered from a results page.


leighsmith:
Anything I can provide to have this pull request accepted?

PR #3670, opened 2017-01-25
Use sigv4 by default for unknown aws s3 regions

JordonPhillips:
Previously we had whitelisted sigv4 regions, this changes that to instead whitelist sigv2 regions. This way, new S3 regions will all use sigv4 without any additional changes needing to be made.

While doing this, I found some dead code relating to ec2 sigv4, which I removed. @jamesls , is there any reason we might need to keep it around? There was no calling code outside of tests, but I suppose somebody could be using it externally.

### DIFF
--- a/boto/auth.py
+++ b/boto/auth.py
@@ -39,7 +39,7 @@ import hmac
 import os
 import posixpath
 
-from boto.compat import urllib, encodebytes, parse_qs_safe
+from boto.compat import urllib, encodebytes, parse_qs_safe, urlparse
 from boto.auth_handler import AuthHandler
 from boto.exception import BotoClientError
 
@@ -51,8 +51,32 @@ except ImportError:
     sha256 = None
 
 
-# Region detection strings to determine if SigV4 should be used
-# by default.
+# Region detection strings to determine if SigV2 should be used
+# by default
+S3_AUTH_DETECT = [
+    '-ap-northeast-1',
+    '.ap-northeast-1',
+    '-ap-southeast-1',
+    '.ap-southeast-1',
+    '-ap-southeast-2',
+    '.ap-southeast-2',
+    '-eu-west-1',
+    '.eu-west-1',
+    '-external-1',
+    '.external-1',
+    '-sa-east-1',
+    '.sa-east-1',
+    '-us-east-1',
+    '.us-east-1',
+    '-us-gov-west-1',
+    '.us-gov-west-1',
+    '-us-west-1',
+    '.us-west-1',
+    '-us-west-2',
+    '.us-west-2'
+]
+
+
 SIGV4_DETECT = [
     '.cn-',
     # In eu-central and ap-northeast-2 we support both host styles for S3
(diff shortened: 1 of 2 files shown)


boto/auth.py
```diff
@@ -1009,27 +1013,6 @@ def get_auth_handler(host, config, provider, requested_capability=None):
     return ready_handlers[-1]
 
 
-def detect_potential_sigv4(func):
```
jamesls: I don't think anything should be using this, but let's not remove it.

You can mark this as deprecated or raise a deprecation warning, but I'm hesitant to remove it outright without any warning.

boto/auth.py
```diff
@@ -51,23 +51,27 @@
...
-    '-ap-south-1',
-    '.us-east-2',
-    '-us-east-2',
-    '-ca-central',
-    '.ca-central',
-    '.eu-west-2',
-    '-eu-west-2',
+# Region detection strings to determine if SigV2 should be used
+# by default
+SIGV2_DETECT = [
```
jamesls: Can we rename this to `S3_AUTH_DETECT` or something along those lines?  This is specifically for the older s3 signature version (which is not sigv2) and we should use consistent naming in boto2, which I believe calls it either `s3` or `hmac-v1`.

boto/auth.py
```diff
@@ -1038,13 +1021,18 @@ def _wrapper(self):
...
-                    return ['hmac-v4-s3']
+        if not hasattr(self, 'host'):
+            return func(self)
+
+        # Use default for the global region and non-aws hosts
+        if 's3.amazonaws.com' in self.host or 'amazonaws.com' not in self.host:
+            return func(self)
+
+        # Use the default for regions that support sigv4 and sigv2
+        if any(test in self.host for test in SIGV2_DETECT):
```
jamesls: How is this code handling the `self.anon` case?

JordonPhillips: While I can handle that case, are there any others like it? Is it a pattern for customers to make their own?

boto/auth.py
JordonPhillips: That's fair

JordonPhillips:
Updated

boto/auth.py
```diff
@@ -1038,13 +1060,22 @@ def _wrapper(self):
...
-            # endpoint-related.
-            for test in SIGV4_DETECT:
-                if test in self.host:
-                    return ['hmac-v4-s3']
+        if not hasattr(self, 'host'):
+            return func(self)
 
-        return func(self)
+        # Use default for the global region and non-aws hosts
+        if 's3.amazonaws.com' in self.host or 'amazonaws.com' not in self.host:
```
kyleknap: We might be missing s3-external as well. How did you generate the ``S3_AUTH_DETECT``? We might want to take a look at the ``endpoints.json`` to make sure none was missed.

kyleknap: The custom host is a good point. I wonder if there is any other way to tell if it is a custom host other than it not containing amazonaws.com?

JordonPhillips: Good point, I think it will be better to parse the url and inspect the netloc. This will also cover any edge cases like the global endpoint much better.

JordonPhillips: see my other comment

JordonPhillips: I generated that using the endpoint resolver to find regions that were compatible with sigv2. It's a good point that `external-1` isn't in there so it should be added.

JordonPhillips:
I've added a lot more robust testing of possible edge cases that didn't already have tests.

JordonPhillips:
Looks like an issue with chinese regions, investigating

boto/auth.py
```diff
@@ -1038,13 +1062,33 @@ def _wrapper(self):
...
+
+        # Use default for non-aws hosts. Adding a url scheme is necessary if
+        # not present for urlparse to properly function.
+        host = self.host
+        if not self.host.startswith('http://') or \
+                self.host.startswith('https://'):
+            host = 'https://' + host
+        netloc = urlparse(host).netloc
+        if not (netloc.endswith('amazonaws.com') or
+                netloc.endswith('amazonaws.com.cn')):
```
JordonPhillips: This is a bit more robust than before in that a customer with a domain like `amazonaws.com.mydomain.com` or somesuch won't hit the sigv4 switch, but now we have to account for every tld that amazon is using. So do we foresee us having more tld's besides those? We haven't done so for any other country.

JordonPhillips:
Just to capture offline discussion: we believe we've found all the edge cases, though a few don't have tests. We'll be moving the tests over to generated tests so that we catch all error in a single run.

JordonPhillips:
`_yield_region_tests` exists because otherwise those same three cases will have to be copied and pasted in three different places. I would rather not break those back out.

kyleknap:
@JordonPhillips that's fine. I was just making suggestions on how to make it easier to read. I think if you break up the test function ``test_s3_signature_version()`` it will be a lot easier to follow what is getting checked.

JordonPhillips:
I broke up the tests and tried to clarify the yielding function.

PR #3673, opened 2017-02-03
Respect query_auth setting when adding x-amz-security-token for S3 URL generation.

### LINKED ISSUE #2043: Generating a S3 url without authentication fails for IAM roles

jperras:
When the S3 component of Boto was used on an EC2 instance that was granted permissions via an IAM role, it attempted to append `x-amz-security-token` key and value to the URL generated via
the `generate_url` method on the available connection, even though `query_auth=False` was explicitly being set.

Includes a simple test case, and the test passes on all interpreters specified in the current tox configuration.

Ready and willing to amend this pull request in whatever fashion is necessary if it is not deemed suitable at this time.

Fixes #2043
Refs. #2044

### DIFF
--- a/boto/s3/connection.py
+++ b/boto/s3/connection.py
@@ -401,8 +401,9 @@ class S3Connection(AWSAuthConnection):
         if response_headers:
             for k, v in response_headers.items():
                 extra_qp.append("%s=%s" % (k, urllib.parse.quote(v)))
-        if self.provider.security_token:
-            headers['x-amz-security-token'] = self.provider.security_token
+        if query_auth:
+            if self.provider.security_token:
+                headers['x-amz-security-token'] = self.provider.security_token
         if extra_qp:
             delimiter = '?' if '?' not in auth_path else '&'
             auth_path += delimiter + '&'.join(extra_qp)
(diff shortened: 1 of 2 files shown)


jperras:
Bump?

Undo1:
Running into this when trying to generate a presigned POST in Lambda (permissions granted through a role).

ervoiox:
Yes this would be helpful for our legacy applications. Please merge.

efimerdlerkravitz:
@Undo1 I have exactly the same problem, what did you do eventually ?

Undo1:
@efimerdlerkravitz I went a different direction and was able to allow public access to the file. It's not a great solution, but in this case it works.

chenli-appD:
I met same issue with Django project. Finally got rid of this issue by updating to boto3. Hope this help.

PR #3676, opened 2017-02-14
Fix host header in S3Connection

cfga:
Fixing a bug in s3.connection.S3Connection.make_request. Host header is being malformed since port is appended to the hostname twice. First, the call self.server_name() in s3.connection.S3Connection.make_request returns hostname with the port already appended, and then this string is passed as a 'host' parameter to the constructor of HTTPRequest in connection.AWSAuthConnection.build_base_http_request.


insequent:
Fixing same bug as https://github.com/boto/boto/pull/3513

Issue #3666, opened 2017-01-09
Bug with uploading to Google Cloud Storage


kenhahn85:
Here is the stack trace.

```
Environment:


Request Method: POST
Request URL: http://localhost:8000/admin/staff/staffprofile/add/

Django Version: 1.10.4
Python Version: 2.7.10
Installed Applications:
['grappelli',
 'nested_admin',
 'django.contrib.admin',
 'django.contrib.admindocs',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'django_jinja',
 'django_select2',
 'concurrency',
 'reversion',
 'reversion_compare',
 'django_js_reverse',
 'django_extensions',
 'root',
 'quote',
 'staff',
 'bootstrapform']
Installed Middleware:
('root.middleware.CustomLocaleMiddleware',
 'django.middleware.security.SecurityMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware')



Traceback:

File "/Users/ken47/.virtualenvs/koram2/lib/python2.7/site-packages/django/core/handlers/exception.py" in inner
  39.             response = get_response(request)

File "/Users/ken47/.virtualenvs/koram2/lib/python2.7/site-packages/django/core/handlers/base.py" in _get_response
  187.                 response = self.process_exception_by_middleware(e, request)

File "/Users/ken47/.virtualenvs/koram2/lib/python2.7/site-packages/django/core/handlers/base.py" in _get_response
  185.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/Users/ken47/.virtualenvs/koram2/lib/python2.7/site-packages/django/contrib/admin/options.py" in wrapper
  544.                 return self.admin_site.admin_view(view)(*args, **kwargs)

File "/Users/ken47/.virtualenvs/koram2/lib/python2.7/site-packages/django/utils/decorators.py" in _wrapped_view
  149.                     response = view_func(request, *args, **kwargs)

[...]

Issue #3665, opened 2017-01-06
Google storage: storage class renamed from STANDARD to MULTI_REGIONAL


stmos:
It looks like newly created buckets on Google storage have the default storage class set to MULTI_REGIONAL instead of STANDARD. This causes issues when trying to save data to the cloud.

I'm using Django 1.8.11, boto 2.45 and django-storages 1.5.1.

This is the part of code that's failing in boto/s3/key.py:

`if self._storage_class not in [None, 'STANDARD']:
        headers[provider.storage_class_header] = self.storage_class`

I was able to "fix" this by using gsutil and forcing the standard class back to 'STANDARD':

`gsutil defstorageclass set STANDARD gs://bucket`

Not sure what would be the best place to fix this.

PR #3678, opened 2017-02-16
Resolve x-goog-storage-class header

peterlandry:
issue #3666
issue #3665

### DIFF
--- a/boto/provider.py
+++ b/boto/provider.py
@@ -155,7 +155,7 @@ class Provider(object):
             # Note that this version header is not to be confused with
             # the Google Cloud Storage 'x-goog-api-version' header.
             VERSION_ID_HEADER_KEY: GOOG_HEADER_PREFIX + 'version-id',
-            STORAGE_CLASS_HEADER_KEY: None,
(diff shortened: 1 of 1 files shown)


atbaker:
Bumped into this problem myself today. Not sure why it just cropped up now, but this fix did the trick for me. I just confirmed it.

Thanks @peterlandry for putting together the PR. +1 on a merge from me.

peterlandry:
It could have just shown up if you created a new bucket. They changed the bucket definitions, so `STANDARD` is no longer an option. Earlier in the pipeline, boto special-cases `STANDARD` and doesn't look for any storage class headers (https://github.com/boto/boto/blob/develop/boto/s3/key.py#L896) . It's only when it sees something else that it tries (and fails to resolve) the header. That's why manually setting the storage class to `STANDARD` "fixes" it.

peterlandry:
And for reference, the new classes: https://cloud.google.com/storage/docs/storage-classes

peterlandry:
Closing, https://github.com/boto/boto/pull/3691 has taken care of this!

PR #3686, opened 2017-02-23
add `iso_date` param in generate_url

Liuchang0812:
It's needed for that we want to take mock testing between Expiration and Skew-Time Error, also, User could uses this param to generate a pre-sign url which is accessable after some time.


Liuchang0812:
jenkins test this please

Liuchang0812:
Hi, please take a look

PR #3691, opened 2017-03-01
Populate storage class from HEAD Object responses.

houglum:
Also fixes two failing ACL tests.

### DIFF
--- a/boto/s3/key.py
+++ b/boto/s3/key.py
@@ -230,6 +230,18 @@ class Key(object):
         else:
             self.encrypted = None
 
+    def handle_storage_class_header(self, resp):
+        provider = self.bucket.connection.provider
+        if provider.storage_class_header:
+            self._storage_class = resp.getheader(
+                provider.storage_class_header, None)
+            if (self._storage_class is None and
+                provider.get_provider_name() == 'aws'):
+                # S3 docs for HEAD object requests say S3 will return this
+                # header for all objects except Standard storage class objects.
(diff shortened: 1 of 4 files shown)


boto/s3/key.py
```diff
@@ -230,6 +230,18 @@ def handle_encryption_headers(self, resp):
...
 
+    def handle_storage_class_header(self, resp):
+        provider = self.bucket.connection.provider
+        if provider.storage_class_header:
+            self._storage_class = resp.getheader(
+                provider.storage_class_header, None)
+            if (self._storage_class is None and
+                provider.get_provider_name() == 'aws'):
+                # S3 docs for HEAD object requests say S3 will return this
+                # header for all objects except Standard storage class objects.
```
mfschwartz: I don't understand why you want to populate the storage class in this case, if the S3 docs say they won't populate the header in this case?

houglum: If the object were of a nondefault class (e.g. STANDARD_IA), this header would be present with that value; only then does S3 add the storage class header to the response.  As such, we can assume it's STANDARD when the header is not present.  This allows us to avoid calling the _get_storage_class method, which performs an additional self.bucket.list call.  Rather than do this (and potentially fail if the user doesn't have list permisson), we can obtain the storage class here.

For context, GSUtil will take advantage of this to populate the object's storage class for the output of `gsutil stat <obj_url>` and `gsutil ls -L <obj_url>`.

Issue #3561, opened 2016-06-21
Boto S3 throws type error in production environment


anshul35:
It is working fine on my local, but when I try to use it in celery, in production environment, it throws a type error. Can someone guide me, what could have caused the error?

```python
 bucket = conn.get_bucket(settings.AWS_STORAGE_BUCKET_NAME)
  File "/usr/local/lib/python3.5/dist-packages/boto/s3/connection.py", line 506, in get_bucket
    return self.head_bucket(bucket_name, headers=headers)
  File "/usr/local/lib/python3.5/dist-packages/boto/s3/connection.py", line 525, in head_bucket
    response = self.make_request('HEAD', bucket_name, headers=headers)
  File "/usr/local/lib/python3.5/dist-packages/boto/s3/connection.py", line 668, in make_request
    retry_handler=retry_handler
  File "/usr/local/lib/python3.5/dist-packages/boto/connection.py", line 1071, in make_request
    retry_handler=retry_handler)
  File "/usr/local/lib/python3.5/dist-packages/boto/connection.py", line 913, in _mexe
    self.is_secure)
  File "/usr/local/lib/python3.5/dist-packages/boto/connection.py", line 705, in get_http_connection
    return self.new_http_connection(host, port, is_secure)
  File "/usr/local/lib/python3.5/dist-packages/boto/connection.py", line 747, in new_http_connection
    connection = self.proxy_ssl(host, is_secure and 443 or 80)
  File "/usr/local/lib/python3.5/dist-packages/boto/connection.py", line 796, in proxy_ssl
    sock.sendall("CONNECT %s HTTP/1.0\r\n" % host)
**TypeError: a bytes-like object is required, not 'str'**
```

TRManderson:
This is a Python 3 compatibility issue. It should work fine if you use Python 2.

Edit 2018: When I made this comment, S3 didn't claim Py3 support

colin-nolan:
For anyone else who encounters this problem and ends up here - there is a merge request with a fix: https://github.com/boto/boto/pull/2718.

ps-jay:
Also note: this issue is only triggered when using proxies.

MajorTal:
...and is there a solution when using proxies?

ps-jay:
Yeah, this need to be merged: https://github.com/boto/boto/pull/3699

PR #3695, opened 2017-03-09
Removes documented support for Python 3 with S3 service

jrandall:
The S3 service does not entirely work with Python 3 (as pointed out in issue #3561). This issue issue has been fixed by PR #2718 long before it was even reported, yet that PR has not been merged yet. Until that PR is merged, the boto documentation should not list Python 3 support for the S3 service, as it does not work in Python 3.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -120,7 +120,7 @@ At the moment, boto supports:
 
 * Storage
 
-  * Amazon Simple Storage Service (S3) (Python 3)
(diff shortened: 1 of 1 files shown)


xiaket:
I'm afraid but #3699 affects all services, it is not limited to S3. Anyone who uses python3/boto and behind a proxy will be affected.

PR #3706, opened 2017-03-27
Do not inject linebreaks into proxy auth header

ffleming:
Using `encodebytes`/`encodestring` injects newlines into the base64 string.  When this is used as a proxy auth header, it breaks.  We see this when the supplied proxy username/password combination is sufficiently long.


98tango:
We've just run into this as well. Can we get this fix merged in please.

PR #3720, opened 2017-04-24
Add space between Authorization header fields

xerebz:
According to the documentation, there are spaces between Credential, SignedHeaders, and Signature fields http://docs.aws.amazon.com/general/latest/gr/sigv4-add-signature-to-request.html#sigv4-add-signature-auth-header.


xerebz:
FWIW, there are spaces between these fields when generated by botocore/boto3/aws-cli.

PR #3721, opened 2017-04-28
boto/ses/connection.py: Added extra functions for notification topics…

github-abcde:
… and mail_from domain.

Still not sure whether 'true' and 'false' strings in return values should be translated to True and False booleans, or whether that should be done in the calling functions.


github-abcde:
Closing due to inactivity

PR #3728, opened 2017-05-09
CA-5976: Disable SSL validation

RevolutionTech:
Since this is broken for buckets with dots in the bucket name.


RevolutionTech:
Sorry, I didn't mean to create this PR. X_X;

PR #3734, opened 2017-05-25
Fix generate_url() AttributeError when using anonymous connections

### LINKED ISSUE #1540: generate_url raises exception for anonymous connections
It would be nice to be able to call generate_url on keys from anonymous connections

Right now if you open a connection anonymously

  conn = boto.connect_s3(anon=True)

Then later try to call generate_url on a key form the bucket, it fails with the following exception.

AttributeError: 'AnonAuthHandler' object has no attribute '_hmac_256'

edmorley:
Previously generating a URL would fail for anonymous connections (even when using `query_auth=False`), with:
`AttributeError: 'AnonAuthHandler' object has no attribute '_hmac_256'`

That now works, and in addition the `query_auth=False` parameter is now optional for anonymous connections.

Fixes #1540.

### DIFF
--- a/tests/unit/s3/test_connection.py
+++ b/tests/unit/s3/test_connection.py
@@ -48,6 +48,18 @@ class TestSignatureAlteration(AWSMockServiceTestCase):
         )
 
 
+class TestAnon(MockServiceWithConfigTestCase):
+    connection_class = S3Connection
+
+    def test_generate_url(self):
+        conn = self.connection_class(
+            anon=True,
+            host='s3.amazonaws.com'
+        )
+        url = conn.generate_url(0, 'GET', bucket='examplebucket', key='test.txt')
+        self.assertNotIn('Signature=', url)
+
+
 class TestPresigned(MockServiceWithConfigTestCase):
     connection_class = S3Connection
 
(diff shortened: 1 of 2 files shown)


tests/unit/s3/test_connection.py
```diff
@@ -48,6 +48,18 @@ def test_switched(self):
         )
 
 
+class TestAnon(MockServiceWithConfigTestCase):
+    connection_class = S3Connection
+
+    def test_generate_url(self):
+        conn = self.connection_class(
+            anon=True,
```
mfschwartz: Is there another test for that exercises that the non-anonymous case generates a URL as expected?

edmorley: Yeah, the existing tests here (since `anon` defaults to `False`):
https://github.com/boto/boto/blob/af045f93d70fbb4cdac2a8e57d040ce59935d45b/tests/unit/s3/test_connection.py#L67-L73
https://github.com/boto/boto/blob/af045f93d70fbb4cdac2a8e57d040ce59935d45b/tests/unit/s3/test_connection.py#L208-L214

mfschwartz: Ok, thanks. LGTM.

edmorley:
Many thanks! 👍

PR #3737, opened 2017-06-02
Use RegionInfo by default with heuristics

JordonPhillips:
The generic `connect` wasn't using the `RegionInfo` class by default. This had been masked by two things:

1. `get_regions` properly uses this as the default. Most tests only go through `get_regions` so most tests wouldn't encounter this at all.
2. The tests explicitly for the region heuristics all provide a custom region class.

This fixes the issue and adds tests that would fail if the default were not correctly set.

Anybody on an impacted version can unblock themselves by passing `region_cls=RegionInfo` into their connect call.

### DIFF
--- a/boto/regioninfo.py
+++ b/boto/regioninfo.py
@@ -205,6 +205,8 @@ def connect(service_name, region_name, region_cls=None,
 
     :returns: A configured connection class.
     """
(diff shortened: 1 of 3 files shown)


xerebz:
LGTM

PR #3738, opened 2017-06-13
Allow specifying s3 host from boto config file.

houglum:
This allows users to specify a value for s3:host in their Boto config file, rather than needing to pass the host directly to the constructor for S3Connection.  There are some situations (e.g. when using gsutil) where users don't have direct control over how the constructor is called; this allows passing the host argument in such scenarios.

(This also fixes an s3 integration test flake.)

### DIFF
--- a/boto/s3/connection.py
+++ b/boto/s3/connection.py
@@ -161,7 +161,7 @@ class HostRequiredError(BotoClientError):
 
 class S3Connection(AWSAuthConnection):
 
-    DefaultHost = boto.config.get('s3', 'host', 's3.amazonaws.com')
+    DefaultHost = 's3.amazonaws.com'
     DefaultCallingFormat = boto.config.get('s3', 'calling_format', 'boto.s3.connection.SubdomainCallingFormat')
     QueryString = 'Signature=%s&Expires=%d&AWSAccessKeyId=%s'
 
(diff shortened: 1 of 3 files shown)


houglum:
@mfschwartz - could you give this a review?

houglum:
Note: This aims to address https://issuetracker.google.com/issues/62161892

boto/s3/connection.py
```diff
@@ -161,7 +161,7 @@ class HostRequiredError(BotoClientError):
 
 class S3Connection(AWSAuthConnection):
 
-    DefaultHost = boto.config.get('s3', 'host', 's3.amazonaws.com')
+    DefaultHost = 's3.amazonaws.com'
```
mfschwartz: Why remove configurability?

boto/s3/connection.py
```diff
@@ -174,9 +174,12 @@ def __init__(self, aws_access_key_id=None, aws_secret_access_key=None,
                  suppress_consec_slashes=True, anon=False,
                  validate_certs=None, profile_name=None):
         no_host_provided = False
+        # Try falling back to the boto config file's value, if present.
         if host is NoHostProvided:
-            no_host_provided = True
-            host = self.DefaultHost
+            host = boto.config.get('s3', 'host')
+            if host is None:
```
mfschwartz: It would be more Pythonic to do:
  if not host:

tests/integration/s3/test_key.py
```diff
@@ -419,7 +419,10 @@ def test_header_encoding(self):
         remote_metadata = check._get_remote_metadata()
 
         # TODO: investigate whether encoding ' ' as '%20' makes sense
-        self.assertEqual(check.cache_control, 'public,%20max-age=500')
+        self.assertIn(
+            check.cache_control,
+            ('public,%20max-age=500', 'public, max-age=500')
+        )
```
mfschwartz: This change looks unrelated to the CL description?

houglum: Yep - didn't mention it in the commit message, but I put a little note about it at the bottom of the pull request.  It's just been an annoying test "failure" to have to look over every time I run the s3 integration tests.

boto/s3/connection.py
houglum: I'm explicitly testing for the NoneType here (which is the default from get() if 'host' isn't found in the boto config).

from our style guide: "Beware of writing if x: when you really mean if x is not None:-e.g., when testing whether a variable or argument that defaults to None was set to some other value. The other value might be a value that's false in a boolean context!"

PR #3748, opened 2017-08-16
Add py35 and py36 to tox config

tedmiston:
While it looks like Python 3.5 and 3.6 are not officially supported the unit tests have [already been updated](https://github.com/boto/boto/blob/master/docs/source/releasenotes/v2.39.0.rst) to pass in 3.5 and already pass in 3.6 as well.

A similar PR https://github.com/boto/boto/pull/2458 was merged to support pypy in tox.

(My current use case is running Apache Airflow which runs on Python through 3.6.x but currently uses boto rather than boto3 behind the scenes.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
   - "3.3"
   - "3.4"
   - "3.5"
+  - "3.6"
   - "pypy"
 env:
   - BOTO_CONFIG=/tmp/nowhere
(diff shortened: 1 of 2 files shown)


tedmiston:
Hey @JordonPhillips @jamesls @kyleknap, I just wanted to tag a few recent heavy committers to see if there's potential interest in merging this small PR?  I know boto3 is more active and that it already has an equivalent patch, but it would nice to have it here as well.

UmanShahzad:
@JordonPhillips Can anyone take a look at this?

tedmiston:
Any chance we could get this merged?

PR #3759, opened 2017-09-13
Revert Billing Configuration Changes

markhjelm:


markhjelm:
Cancelling this pull request because it doesn't fix the as-it-turns-out unrelated unit test problem with tests/unit/glacier/test_writer.py.

PR #3765, opened 2017-10-04
Update layer1.py

knil-sama:
Add more information on rebuild_environment

Get this error on aws
@mfschwartz 
An error occurred (InvalidParameterValue) when calling the RebuildEnvironment operation: No Environment found for EnvironmentName = 'XXXXXXX'.

### DIFF
--- a/boto/beanstalk/layer1.py
+++ b/boto/beanstalk/layer1.py
@@ -772,9 +772,8 @@ class Layer1(AWSQueryConnection):
             Condition: You must specify either this or an EnvironmentId, or
             both.  If you do not specify either, AWS Elastic Beanstalk returns
             MissingRequiredParameter error.
-            If the environment hadn't be created or is terminated returns 
-            InvalidParameterValue error.
 
+        :raises InvalidParameterValue: If the environment hadn't be created or is terminated
         :raises: InsufficientPrivilegesException
         """
         params = {}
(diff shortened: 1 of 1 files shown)


boto/beanstalk/layer1.py
```diff
@@ -772,6 +772,8 @@ def rebuild_environment(self, environment_id=None, environment_name=None):
             Condition: You must specify either this or an EnvironmentId, or
             both.  If you do not specify either, AWS Elastic Beanstalk returns
             MissingRequiredParameter error.
+            If the environment hadn't be created or is terminated returns
```
mfschwartz: Can you please clarify why this might happen? What could cause environment creation to fail?

knil-sama: I just want to clarify that the environment must already exist and be in a not terminated state.
My use case is that I wanted to use it to rebuild in the morning environment I terminated in the night, that raise this InvalidParameterValue that wasn't documented.

mfschwartz: (FYI I haven't worked with this part of the code, which is why I'm asking these questions.)

The thing that confused me is that the function is called "rebuild_environment", so at first I thought your comment was saying the function wasn't able to rebuild the environment, rather than there was a problem with the environment you passed in.

How about this:
  - Move this to a raises: clause below
  - In that clause say "InvalidParameterValue if environment_name doesn't refer to a currently active environment."

knil-sama:
@mfschwartz Yes that would be better

PR #3770, opened 2017-10-25
Use pkg_resources to keep the package zip-safe

frivoire:
This resolves a bug in the packaging of this package that occurs in some situation :

When `boto` is installed manually from a source distribution, in a `PYTHONDONTWRITEBYTECODE=1` environment, it ends up in the shape of a **zipfile egg** in dist-packages, as it is not marked **zip-unsafe**.
This means that importing it will result in an error, because `regioninfo.py / _load_json_file()` will try to open `endpoints.json` in what it thinks is inside a directory, when it is actually inside a zipfile.

The package could be marked as **zip-unsafe**, but there is an even better way to solve the problem: `pkg_resources` (which is part of setuptools) that will take care (when needed) of unzipping and caching the data file, returning a valid filename which can be consumed.

### DIFF
--- a/boto/__init__.py
+++ b/boto/__init__.py
@@ -26,6 +26,7 @@
 #
 from boto.pyami.config import Config, BotoConfigLocations
 from boto.storage_uri import BucketStorageUri, FileStorageUri
+from pkg_resources import resource_filename
 import boto.plugin
 import datetime
 import os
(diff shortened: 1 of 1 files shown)


frivoire:
Any remarks on this PR ?
Thanks !

frivoire:
Dont hesitate if any question/remark on this PR.

frivoire:
Any news on this PR ?
Thanks for your help :)

PR #3776, opened 2017-11-09
Add default region to determine_region_name

gaul:
Non-Amazon hosts do not match any of the conditionals which previously
raised an `UnboundLocalError`.


gaul:
Spurious test failure in `pip install` for one of the matrix builds.

gaul:
Anything blocking this commit or can someone suggest a workaround?

PR #3777, opened 2017-11-14
Add documentation about S3 in boto configuration docs.

UmanShahzad:
I've come across the use case of applying the `host` var in a hitherto undocumented `s3` section in the boto configuration file documentation. I've added that and any other variables used under the `s3` section throughout the codebase.


UmanShahzad:
Tests are failing for unrelated reasons.

Issue #1791, opened 2013-10-14
Fix to s3.Key.get_contents_to_filename for cases where an exception is thrown


nmerket:
I was testing some code where I wanted it to recover gracefully in the case of an intermittent internet connection. I noticed in the get_contents_to_filename method it was returning an error related to trying to delete a file that was open. Upon looking into the file, I noticed that in the case an exception was thrown, it tried to os.remove the file before closing it. It's pretty easy to solve with a with statement. Here's a quick patch that solves the problem. 

```
Index: key.py
===================================================================
--- key.py  (revision 3608)
+++ key.py  (working copy)
@@ -1623,17 +1623,15 @@
             with the stored object in the response.  See
             http://goo.gl/EWOPb for details.
         """
-        fp = open(filename, 'wb')
         try:
-            self.get_contents_to_file(fp, headers, cb, num_cb, torrent=torrent,
-                                      version_id=version_id,
-                                      res_download_handler=res_download_handler,
-                                      response_headers=response_headers)
+            with open(filename, 'wb') as fp:
+                self.get_contents_to_file(fp, headers, cb, num_cb, torrent=torrent,
+                                          version_id=version_id,
+                                          res_download_handler=res_download_handler,
+                                          response_headers=response_headers)
         except Exception:
             os.remove(filename)
             raise
-        finally:
-            fp.close()
         # if last_modified date was sent from s3, try to set file's timestamp
         if self.last_modified != None:
             try:

```

I know this probably isn't the right way to submit a fix, but it's a fairly simple one and I wanted to make sure you got it. 

Thanks.

PR #3790, opened 2018-02-19
Fix get_contents_to_filename OSError when filename is a directory

benlk:
## Changes

In `boto/s3/key.py`, `get_contents_to_filename`,  when `os.remove(filename)` fails, this tries again with os.rmdir(filename) in the off chance that a directory is being attempted to be removed.

## Why

`os.remove(filename)` raises an OSError when the filename to be removed is that of a directory. 

I think this is also what https://github.com/boto/boto/issues/1791 was attempting to fix.

For testing, here's the code that was calling s3.get_contents_to_filename: https://github.com/INN/power-players/blob/7efb569/fabfile/assets.py#L219-L230 was called by https://github.com/INN/power-players/blob/7efb569/fabfile/assets.py#L110 as a [Fabric](http://www.fabfile.org/) command.

### DIFF
--- a/boto/s3/key.py
+++ b/boto/s3/key.py
@@ -1723,8 +1723,16 @@ class Key(object):
                                           res_download_handler=res_download_handler,
                                           response_headers=response_headers)
         except Exception:
-            os.remove(filename)
-            raise
+            try:
+                os.remove(filename)
+            except OSError:
+                try:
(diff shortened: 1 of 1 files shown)


benlk:
Closing this because the described behavior actually was revealing an issue in the tool I was using, not boto.

PR #3799, opened 2018-03-08
Support fetching GCS bucket encryption metadata.

houglum:
Also fixes a broken integration test in tests/integration/s3/test_key.py.

### DIFF
--- a/boto/exception.py
+++ b/boto/exception.py
@@ -475,9 +475,12 @@ class InvalidCorsError(Exception):
         self.message = message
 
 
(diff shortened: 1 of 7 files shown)


houglum:
Looks like the CI test failures are due to problems with the "cryptography" PyPI package (they don't support Python 2.6 anymore, and the Python 3.3 pip installation of that package failed), rather than this small change.  Py2.7, Py3.4+, and PyPy environments all passed.

houglum:
@mfschwartz, could you merge this when you get the chance?

PR #3800, opened 2018-03-12
Fix to support uploads to KMS-encrypted buckets.

houglum:
This fixes uploads to buckets with default KMS keys failing, e.g.:

$ echo "<some text here>" | ./gsutil cp - gs://<MyKmsEncryptedBucketHere>/object1
GSDataError: BotoClientError: ETag from S3 did not match computed MD5. "-CMPujcLB4NkCEAE=" vs. 25a14f7d6fdde4161404a71d7817a0b4

### DIFF
--- a/boto/s3/key.py
+++ b/boto/s3/key.py
@@ -236,7 +236,7 @@ class Key(object):
             self._storage_class = resp.getheader(
                 provider.storage_class_header, None)
             if (self._storage_class is None and
-                provider.get_provider_name() == 'aws'):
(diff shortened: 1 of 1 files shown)


houglum:
@mfschwartz, please merge when you get the chance. As with the previous pull request, the failing TravisCI tests are due to the cryptography package (in Py 2.6 and 3.3) that Boto depends on.  I've run all the integration tests and manually tested with gsutil as well.  Thanks!

PR #3801, opened 2018-03-14
Fix AWS V2 auth with query params and security token.

khasanovbi:
S3 use SecurityToken query arg while use v2 auth with security token.
https://docs.aws.amazon.com/general/latest/gr/signature-version-2.html#calculating-signature-2


KhasanovBI:
Is anyone alive here?)

PR #3817, opened 2018-07-11
Don't append 443 to host in server_name in PY3

i-trofimtschuk:
On python3 the branch for older broken version <2.6 was used


i-trofimtschuk:
our project moved on to using boto3

PR #3820, opened 2018-07-27
Python3 gs compatibility

edhodapp:
This PR passes all of the s3 and gs tests under Python 2.7 and Python 3.4.

### DIFF
--- a/boto/gs/bucket.py
+++ b/boto/gs/bucket.py
@@ -453,8 +453,8 @@ class Bucket(S3Bucket):
             headers['x-goog-if-metageneration-match'] = str(if_metageneration)
 
         response = self.connection.make_request(
-            'PUT', get_utf8_value(self.name), get_utf8_value(key_name),
-            data=get_utf8_value(data), headers=headers, query_args=query_args)
+            'PUT', self.name, key_name,
+            data=data, headers=headers, query_args=query_args)
         body = response.read()
         if response.status != 200:
             raise self.connection.provider.storage_response_error(
@@ -599,7 +599,7 @@ class Bucket(S3Bucket):
         :param dict headers: Additional headers to send with the request.
         """
         response = self.connection.make_request(
-            'PUT', get_utf8_value(self.name), data=get_utf8_value(cors),
+            'PUT', self.name, data=cors,
             query_args=CORS_ARG, headers=headers)
         body = response.read()
         if response.status != 200:
@@ -1004,7 +1004,7 @@ class Bucket(S3Bucket):
         """
         xml = lifecycle_config.to_xml()
         response = self.connection.make_request(
-            'PUT', get_utf8_value(self.name), data=get_utf8_value(xml),
+            'PUT', self.name, data=xml,
             query_args=LIFECYCLE_ARG, headers=headers)
         body = response.read()
         if response.status == 200:
@@ -1126,7 +1126,7 @@ class Bucket(S3Bucket):
         body = self._construct_encryption_config_xml(
             default_kms_key_name=default_kms_key_name)
         response = self.connection.make_request(
-            'PUT', get_utf8_value(self.name), data=get_utf8_value(body),
+            'PUT', self.name, data=body,
             query_args=ENCRYPTION_CONFIG_ARG, headers=headers)
         body = response.read()
         if response.status != 200:
--- a/boto/gs/bucket.py
+++ b/boto/gs/bucket.py
@@ -19,6 +19,10 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
 # IN THE SOFTWARE.
 
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
 import re
 import urllib
 import xml.sax
--- a/boto/utils.py
+++ b/boto/utils.py
@@ -864,7 +864,7 @@ def get_utf8_value(value):
         return value
 
     if not isinstance(value, six.string_types):
-        value = six.text_type(value)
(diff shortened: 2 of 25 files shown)


houglum:
Heads up @mfschwartz -- I'm going to give this first pass for review, and I'll ping you when we're ready for you to take a pass and eventually merge it.

boto/gs/bucket.py
```diff
@@ -453,8 +457,8 @@ def _set_acl_helper(self, acl_or_str, key_name, headers, query_args,
             headers['x-goog-if-metageneration-match'] = str(if_metageneration)
 
         response = self.connection.make_request(
-            'PUT', get_utf8_value(self.name), get_utf8_value(key_name),
-            data=get_utf8_value(data), headers=headers, query_args=query_args)
+            'PUT', self.name, key_name,
```
houglum, line 460: The HTTP request should probably accept bytes (str in Python2, bytes in Python3).  I see two options here:

- If we're going to remove this wrapper from all the call sites, I'd propose invoking `get_utf8_value` on each of the arguments within the method body of `make_connection` in connection.py.  Note that in the event we need to send a set of bytes encoded in some other encoding, this would be fine -- `get_utf8_value` only ensures that the type is bytes, not that they were encoded in UTF8 coding.

- Or, go back to requiring that each caller passes bytes (not unicode) to `make_request`.

houglum: Actually, after talking this over with Ed offline, it looks like the s3 components removed these utf8 wrappers from their `make_connection` calls.

houglum: Ah, I see. The get_utf8_value wrappers were added ~5 years ago [1], but the code was refactored roughly a year later [2] to prevent the issues from mixed unicode/bytes described in [1].

[1] https://github.com/boto/boto/commit/8f466ebcac5fa22743bae37ef3d3312c79348a9b
[2] https://github.com/boto/boto/commit/96cd2800b6db067e503ef499486adf50ee9ca928

boto/utils.py
```diff
@@ -864,7 +864,7 @@ def get_utf8_value(value):
         return value
 
     if not isinstance(value, six.string_types):
-        value = six.text_type(value)
+        value = six.text_type(value).encode('utf-8')
```
houglum, line 867: I don't think this change is necessary.  If we cast `value` to type `six.text_type`, then the next if statement should evaluate to True and do the encoding for us.

tests/integration/gs/test_basic.py
```diff
@@ -97,6 +103,66 @@
   '\s*</AccessControlList>\s*')
 
 
+class XmlCompare(object):
+    """Compares to XML etrees
```
houglum, line 107: Wording/formatting nit -- change to:

    Compares two XML `ElementTree`s.

tests/integration/gs/test_basic.py
```diff
@@ -97,6 +103,66 @@
...
+
+       Modified from this stackoverflow code to properly recurse for misordered
+       elements.
+       https://stackoverflow.com/questions/24492895/comparing-two-xml-files-in-python
+    """
+    def xml_compare(self, tree1, tree2, excludes=[]):
+        """
+        Compares two xml etrees
+        :param tree1: the first tree
```
houglum, line 116: Could you add the types here (and in the method below)?  Looking at the docstring comment style in s3/key.py, this would look something like:

    :type tree1: xml.etree.ElementTree
    :param tree1: <...>

    :type tree2: <...>
    <...>

tests/integration/gs/test_basic.py
```diff
@@ -97,6 +103,66 @@
...
+        :param tree2: the second tree
+        :param excludes: list of string of attributes to exclude from comparison
+        :return:
+            True if both files match
```
houglum, line 120: This can be on the same line as `:return:`.

tests/integration/gs/test_basic.py
```diff
@@ -441,7 +507,9 @@ def test_lifecycle_config_bucket(self):
             LIFECYCLE_CONDITIONS_FOR_SET_STORAGE_CLASS_RULE)
         bucket.configure_lifecycle(lifecycle_config)
         xml = bucket.get_lifecycle_config().to_xml()
-        self.assertEqual(xml, LIFECYCLE_DOC)
+        self.assertTrue(
+            XmlCompare().xml_compare(ET.fromstring(xml),
```
houglum, line 511: It seems unnecessary to need to instantiate an `XmlCompare` object to do this.  I'd suggest either making the methods in XmlCompare `@staticmethod`s, or moving them to the their own module and making them module-level methods (probably cleaner, since that would allow re-using those functions in other modules).

tests/integration/gs/test_basic.py
```diff
@@ -97,6 +103,66 @@
   '\s*</AccessControlList>\s*')
 
 
+class XmlCompare(object):
```
houglum, line 106: [no action required]

Neat! Thanks for adding this in.

houglum:
As an FYI, we're holding off on further review here until Ed's got these changes working in his fork of gsutil and all the unit tests pass in Py 2 and 3.  This involves some other work as well (not related to Boto specifically), but it's worth getting all that work done before we merge anything, as we may potentially find we need to make additional changes in Boto along the way.

edhodapp:
Okay, so the work on gsutil itself made it apparent that:
a) The boto tests are not comprehensive enough for Python 3.
b) That the gsutil tests uncovered holes in the boto coverage that makes it very unlikely that this PR is viable.
I am checking the changes I made for gsutil tests to pass into my fork of boto, and killing this PR for now.

PR #3832, opened 2018-09-10
SWF Layer1 client: Fix argument position error for security token.

mookerji:
- Fixes issue where the SWF client has an argument type error that
  prevents the swf.Layer1 object from being used with token-based
  authentication (i.e,. anything passing AWS_SECURITY_TOKEN). This
  explicitly passes the security_token argument to AWSAuthConnection
  as a keyword, instead of a positional argument.
- swf.Layer1 constructor API change: session_token => security_token,
  as session_token is not an actual attribute of AWSAuthConnection.


mookerji:
Ooof, just noticed that this was a duplicate of https://github.com/boto/boto/pull/1871/files. I can add a unit test and would love to help get this merged.

PR #3833, opened 2018-09-17
Allows anon auth'n to S3 via a config option.

houglum:
See https://stackoverflow.com/questions/52308440/fetch-content-in-aws-s3-public-bucket-from-gcp-data-storage/52322861#52322861
for a valid use case for this option.

This option was named after the "--no-sign-request" option in the AWS
CLI.

The additional changes to provider.py and auth.py are to prevent Boto
from checking and failing to find credentials on the metadata server
(and thus spewing error-level logs to the console), and to allow
pickling to work for the AnonAuthHandler and HmacKeys classes.

### DIFF
--- a/boto/auth.py
+++ b/boto/auth.py
@@ -98,21 +98,26 @@ SIGV4_DETECT = [
 class HmacKeys(object):
     """Key based Auth handler helper."""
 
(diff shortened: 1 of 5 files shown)


houglum:
\+ @mfschwartz for review

PR #3843, opened 2019-01-16
Send SNI for HTTPS connections (required for Google Cloud Storage)

kevinoid:
Several Google servers, including storage.googleapis.com, send an invalid certificate in response to HTTPS connections which do not include Server Name Indication (SNI) in ClientHello:

    OU = "No SNI provided; please fix your client.", CN = invalid2.invalid

This can be demonstrated by running `openssl s_client -connect storage.googleapis.com:443` with and without the `-noservername` option.  This causes errors in boto, such as:

      File "./gsbototest.py", line 6, in <module>
        boto.storage_uri('bucket-name', 'gs').create_bucket()
      File "/tmp/boto/boto/storage_uri.py", line 574, in create_bucket
        storage_class)
      File "/tmp/boto/boto/gs/connection.py", line 95, in create_bucket
        data=get_utf8_value(data))
      File "/tmp/boto/boto/s3/connection.py", line 682, in make_request
        retry_handler=retry_handler
      File "/tmp/boto/boto/connection.py", line 1074, in make_request
        retry_handler=retry_handler)
      File "/tmp/boto/boto/connection.py", line 1033, in _mexe
        raise ex
    ssl.SSLError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:727)

Where `gsbototest.py` has the following content:

```python
#!/usr/bin/env python

import boto
import gcs_oauth2_boto_plugin

boto.storage_uri('bucket-name', 'gs').create_bucket()
```

To fix the issue, this PR sends SNI, where possible.  This requires calling `wrap_socket` on an `SSLContext` instance.  The necessary `SSLContext` is constructed and additional `SSLSocket` attributes are set using the [same code as `SSLSocket`](https://github.com/python/cpython/blob/v2.7.15/Lib/ssl.py#L555-L570) to minimize potential differences introduced.  The code is only called when `SSLContext` is available (Python 2.7.9 and later) and only when OpenSSL is compiled with SNI support.

Thanks for considering,
Kevin

### DIFF
--- a/boto/https_connection.py
+++ b/boto/https_connection.py
@@ -125,10 +125,28 @@ class CertValidatingHTTPSConnection(http_client.HTTPConnection):
         else:
             msg += "using system provided SSL certs"
         boto.log.debug(msg)
-        self.sock = ssl.wrap_socket(sock, keyfile=self.key_file,
-                                    certfile=self.cert_file,
-                                    cert_reqs=ssl.CERT_REQUIRED,
-                                    ca_certs=self.ca_certs)
+        if hasattr(ssl, 'SSLContext') and getattr(ssl, 'HAS_SNI', False):
+            # Use SSLContext so we can specify server_hostname for SNI
+            # (Required for connections to storage.googleapis.com)
+            context = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
+            context.verify_mode = ssl.CERT_REQUIRED
+            if self.ca_certs:
+                context.load_verify_locations(self.ca_certs)
+            if self.cert_file:
+                context.load_cert_chain(self.cert_file, self.key_file)
+            self.sock = context.wrap_socket(sock, server_hostname=self.host)
+            # Add attributes only set in SSLSocket constructor without context:
+            self.sock.keyfile = self.key_file
+            self.sock.certfile = self.cert_file
+            self.sock.cert_reqs = context.verify_mode
+            self.sock.ssl_version = ssl.PROTOCOL_SSLv23
+            self.sock.ca_certs = self.ca_certs
+            self.sock.ciphers = None
+        else:
+            self.sock = ssl.wrap_socket(sock, keyfile=self.key_file,
+                                        certfile=self.cert_file,
+                                        cert_reqs=ssl.CERT_REQUIRED,
+                                        ca_certs=self.ca_certs)
         cert = self.sock.getpeercert()
         hostname = self.host.split(':', 0)[0]
         if not ValidateCertificateHostname(cert, hostname):


houglum:
Given the note about `-noservername`, I found that this only happens when the version of OpenSSL your Python installation is built against is 1.1.1 -- 1.1.0 and below doesn't have that flag, and this isn't a problem for people using versions prior to 1.1.1.

Most GCE VMs I've created have used version 1.0.1, 1.0.2, or 1.1.0 of OpenSSL, which explains why most folks aren't seeing this.   <strike>Looking at the OpenSSL downloads page, 1.1.1 is currently in alpha, so I'm not sure if this new behavior will change or not.  I'd opt for waiting until this becomes part of the stable version before we pull in any fixes to address it.</strike>

----------------------------------------------------
To reproduce this on an Ubuntu 16 machine, I had to download OpenSSL from source, install pyenv, and build a Python version (I chose 2.7.13) against the newly-installed OpenSSL libraries.  I followed a mix of instructions from [this post](https://websiteforstudents.com/manually-install-the-latest-openssl-toolkit-on-ubuntu-16-04-18-04-lts/), to configure/make OpenSSL 1.1.1, and [this StackOverflow post](https://stackoverflow.com/questions/18752409/updating-openssl-in-python-2-7), to install pyenv and build Python against the new version.

```shell
# Install OpenSSL 1.1.1a:
$ cd /tmp
$ wget https://www.openssl.org/source/openssl-1.1.1.tar.gz
$ tar xvf openssl-1.1.1.tar.gz
$ cd openssl-1.1.1/
$ sudo apt-get install make gcc
$ sudo ./config -Wl,--enable-new-dtags,-rpath,'$(LIBRPATH)'
$ sudo make
$ sudo make install

# Install pyenv:
$ git clone https://github.com/pyenv/pyenv.git ~/.pyenv
$ echo 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.bash_profile
$ echo 'export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.bash_profile
$ echo -e 'if command -v pyenv 1>/dev/null 2>&1; then\n  eval "$(pyenv init -)"\nfi' >> ~/.bash_profile
$ exec "$SHELL"

# Build a new Python 2.7.13 installation against OpenSSL 1.1.1a:
$ export CONFIGURE_OPTS="CPPFLAGS=-I/usr/local/include LDFLAGS=-L/usr/local/lib"
$ sudo apt-get install -y make build-essential libssl-dev zlib1g-dev libbz2-dev libreadline-dev libsqlite3-dev wget curl llvm libncurses5-dev libncursesw5-dev xz-utils tk-dev libffi-dev liblzma-dev python-openssl
$ pyenv 2.7.13
```

And finally, I used the interpreter at ~/.pyenv/versions/2.7.13/bin/python2 to run @kevinoid's repro.  It takes a while to get that error message if your boto config file's `num_retries` isn't set to something low like 1.

kevinoid:
Great analysis!  Thanks @houglum!  That's good information and explains why this issue isn't more widespread.

To clarify, [1.1.1 was released in November](https://mta.openssl.org/pipermail/openssl-announce/2018-November/000141.html).  It's no longer alpha.  OpenSSL uses a confusing versioning scheme with a letter at the end, so 1.1.1a is the first of the 1.1.1 series of releases.

The reason I encountered the issue is that 1.1.1a is currently in Debian Testing (since 2018-11-25).  I assume 1.1.1 will be in Buster when it releases in a few months.

Thanks again for looking into it!  Let me know if we need to figure out why the issue occurs on 1.1.1 and not 1.1.0 more specifically.

kevinoid:
Ha!  I even got the OpenSSL versioning wrong myself.  [1.1.1 was released in September](https://mta.openssl.org/pipermail/openssl-announce/2018-September/000133.html) and [1.1.1a in November](https://mta.openssl.org/pipermail/openssl-announce/2018-November/000141.html) as the second release of the series.  Confusing indeed.

Also note that since `1.1.1~~pre6-1` (including the current version `1.1.1a-1`) Debian compiles OpenSSL with a default security level of 2 (instead of 1 which is the upstream default).  I don't think it's relevant here, since you were able to reproduce the issue with the upstream default, but it has bitten me before so I thought it was worth mentioning.

houglum:
Ah, you're right on the versioning semantics -- thanks.  That makes this a bit higher priority :)

xnox:
BTW this was proposed before already in https://github.com/boto/boto/pull/3836/files as well. But the code there looks different.

Which one is better?

kevinoid:
Good catch @xnox!  I'm not sure how I missed it.  Here are the differences I noticed:

* In #3836, HTTPS connections will fail on Python < 2.7.9 and Python3 < 3.4 due to missing [`ssl.create_default_context`](https://docs.python.org/3/library/ssl.html#ssl.create_default_context) (which caused the [Travis CI build failure on Python 3.3](https://travis-ci.org/boto/boto/jobs/437349552)).
* #3836 adds SNI to `AWSAuthConnection.proxy_ssl`.  #3843 does not.  I am not aware of any issues with AWS caused by lack of SNI, but it may be preferable to send SNI for consistency or future-proofing.
* #3836 replaces [`ssl.wrap_socket`](https://docs.python.org/3/library/ssl.html#ssl.wrap_socket) with [`SSLContext.wrap_socket`](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.wrap_socket) unconditionally, whereas #3843 continues to use [`ssl.wrap_socket`](https://docs.python.org/3/library/ssl.html#ssl.wrap_socket) when SNI support is not available.  I did this to change the code-path only when required.  However, after re-examining it, I think reducing the number of code paths and avoiding a deprecated method is probably a better choice.
* #3836 instantiates the `SSLContext` using [`ssl.create_default_context`](https://docs.python.org/3/library/ssl.html#ssl.create_default_context) while #3843 calls the constructor.  The practical difference is that #3836 calls [`SSLContext.load_default_certs(Purpose.SERVER_AUTH)`](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_default_certs) on the context (before loading `key_file` and/or `ca_certs`) while #3843 never calls it (which matches [`ssl.wrap_socket`](https://docs.python.org/3/library/ssl.html#ssl.wrap_socket)).
* #3843 adds the [attributes set by the `SSLSocket` constructor](https://github.com/python/cpython/blob/v2.7.15/Lib/ssl.py#L555-L570) when called by `ssl.wrap_socket` so `sock` will have the same attributes.  Since these properties don't appear to be present in Python 3, this is probably unnecessary, but again I was trying to minimize changes.

Let me know if I should change the behavior in this PR for any of the above points, or feel free to close and merge #3836.

kevinoid:
> * #3836 replaces [`ssl.wrap_socket`](https://docs.python.org/3/library/ssl.html#ssl.wrap_socket) with [`SSLContext.wrap_socket`](https://docs.python.org/3/library/ssl.html#ssl.SSLContext.wrap_socket) unconditionally, whereas #3843 continues to use [`ssl.wrap_socket`](https://docs.python.org/3/library/ssl.html#ssl.wrap_socket) when SNI support is not available.  I did this to change the code-path only when required.  However, after re-examining it, I think reducing the number of code paths and avoiding a deprecated method is probably a better choice.

After further consideration, this also depends on whether Python without `ssl.SSLContext` should be supported (Python < 2.7.9, Python3 < 3.2).

houglum:
In the interest of not breaking backward compatibility (for anyone that might be using this on old Python versions), I'd prefer this PR, which falls back to the old behavior if SNI isn't available.

Looks good to me.  @mfschwartz, could you review and merge this into Boto?  I've pulled @kevinoid's `add-sni` branch into a local clone of gsutil and run all of our unit and integration tests with it, against both s3:// and gs:// resources.

boto/https_connection.py
```diff
@@ -125,10 +125,28 @@ def connect(self):
         else:
             msg += "using system provided SSL certs"
         boto.log.debug(msg)
-        self.sock = ssl.wrap_socket(sock, keyfile=self.key_file,
-                                    certfile=self.cert_file,
-                                    cert_reqs=ssl.CERT_REQUIRED,
-                                    ca_certs=self.ca_certs)
+        if hasattr(ssl, 'SSLContext') and getattr(ssl, 'HAS_SNI', False):
+            # Use SSLContext so we can specify server_hostname for SNI
+            # (Required for connections to storage.googleapis.com)
```
mfschwartz, line 130: Have you tested what this does to requests sent to S3?

I want to make sure changes made for the sake of GCS functionality don't break S3 customer calls.

houglum: Yep.  I ran the integration tests against S3; the tests were run with v1.1.0 of OpenSSL, which, according to the interpreter session below, should hit this code path:

```
>>> import ssl
>>> hasattr(ssl, 'SSLContext')
True
>>> getattr(ssl, 'HAS_SNI', False)
True
>>> ssl._OPENSSL_API_VERSION
(1, 1, 0, 6, 15)
```

mfschwartz (approved):

mfschwartz:
Why are there Travis failures? I'm ready to merge this PR once they're fixed.

kevinoid:
Thanks @mfschwartz!

Travis failed for Python 2.6 during `pip install -r requirements-py26.txt` with `ImportError: cannot import name check_output` from `from subprocess import check_output` in `.eggs/pycparser-2.19-py2.6.egg/pycparser/__init__.py`.  It appears to be unrelated to this PR and occurs for all recent builds that I checked.

kevinoid:
Thanks for merging it @mfschwartz (and to @houglum and @xnox for helping to review)!

Also, I just submitted #3844 to fix the Python 2.6 build errors.

xnox:
Thank you all!

PR #3846, opened 2019-03-07
adding us-gov-east-1

jayninja:


jayninja:
failed due to py2.6 dependencies. Tests should pass once #3844 is merged.

jered-miller:
+1 Need us-gov-east-1 support added

amkuchta:
Also +1 - this is several months old, any chance of getting it integrated in the near future?

jonneyj:
still no word on when govcloud east will be put into the code base by default?  If not is there a endpoint.json that has all of the appropriate lines of code in it for people to use.  My searches are not proving fruitful and a copy of this code is not working all the way.  It shows up as an option when I input it into my clustered tower instances but then goes away as an option after I save the source.

PR #3847, opened 2019-03-08
Fix HostConnectionPool quadratic runtime

carsonip:
list.pop(0) is O(N) such that get() is quadratic. Use deque and
six.moves.range to keep get() always O(N). Add tests to
HostConnectionPool.

### DIFF
--- a/boto/connection.py
+++ b/boto/connection.py
@@ -52,6 +52,7 @@ import sys
 import time
 import xml.sax
 import copy
+from collections import deque
 
 from boto import auth
 from boto import auth_handler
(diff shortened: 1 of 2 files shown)


carsonip:
Test fails due to the previous SNI commit.

carsonip:
Benchmark script for this commit:
```
import time
from boto.connection import HostConnectionPool


def f():
    p = HostConnectionPool()
    count = 10000
    for _ in xrange(count):
        p.put(None)
    for _ in xrange(count):
        assert p.size() == count - _
        p.get()


t = time.time()
f()
print(time.time() - t)
```

Before fix: 0.27458691597
After fix: 0.0182430744171

jayninja:
> Test fails due to the previous SNI commit.

@carsonip looks like we are waiting on #3844 to be merged to fix python2.6 due to dependency issues :(

PR #3851, opened 2019-03-18
Python 2 / 3 compatibility changes [MOVED: See PR #3856]

catleeball:
@walkerjoe Ed (todo: find Ed's github alias) and I made several fixes to help boto be cross-compatible with Python 2 and 3. These issues were identified while working on making [gsutil's py-six-current branch](https://github.com/GoogleCloudPlatform/gsutil/tree/py-six-current) compatible with Python 2.7 and 3.5+. This was done under the guidance of @houglum

Notably, we're using the [six](https://github.com/benjaminp/six) library to help ensure byte strings vs unicode strings work together nicely. Many of the changes we made revolve around making Python 2 and Python 3's byte vs unicode strings play together nicely, along with changes in division. The ttyprint and ttywrite methods were added to allow type agnostic string printing.

Let us know if you have any questions about these changes! :slightly_smiling_face:

### DIFF
--- a/boto/gs/bucket.py
+++ b/boto/gs/bucket.py
@@ -24,8 +24,10 @@ from __future__ import division
 from __future__ import print_function
 
 import re
-import urllib
 import xml.sax
+
+from six import PY3
+from six.moves.urllib.parse import quote
 
 import boto
 from boto import handler
@@ -52,7 +54,7 @@ ENCRYPTION_CONFIG_ARG = 'encryptionConfig'
 LIFECYCLE_ARG = 'lifecycle'
 STORAGE_CLASS_ARG='storageClass'
 ERROR_DETAILS_REGEX_STR = r'<Details>(?P<details>.*)</Details>'
-if six.PY3:
+if PY3:
     ERROR_DETAILS_REGEX_STR = ERROR_DETAILS_REGEX_STR.encode('ascii')
 ERROR_DETAILS_REGEX = re.compile(ERROR_DETAILS_REGEX_STR)
 
@@ -125,7 +127,7 @@ class Bucket(S3Bucket):
             query_args_l.append('generation=%s' % generation)
         if response_headers:
             for rk, rv in six.iteritems(response_headers):
-                query_args_l.append('%s=%s' % (rk, urllib.quote(rv)))
+                query_args_l.append('%s=%s' % (rk, quote(rv)))
         try:
             key, resp = self._get_key_internal(key_name, headers,
                                                query_args_l=query_args_l)
@@ -361,7 +363,7 @@ class Bucket(S3Bucket):
                     details = (('<Details>%s. Note that Full Control access'
                                 ' is required to access ACLs.</Details>') %
                                details)
-                    if six.PY3:
+                    if PY3:
                         details = details.encode('utf-8')
                     body = re.sub(ERROR_DETAILS_REGEX, details, body)
             raise self.connection.provider.storage_response_error(
--- a/boto/gs/bucket.py
+++ b/boto/gs/bucket.py
@@ -41,9 +41,8 @@ from boto.gs.key import Key as GSKey
 from boto.s3.acl import Policy
 from boto.s3.bucket import Bucket as S3Bucket
 from boto.utils import get_utf8_value
+from boto.compat import quote
 from boto.compat import six
-
-quote = six.moves.urllib.parse.quote
 
 # constants for http query args
 DEF_OBJ_ACL = 'defaultObjectAcl'
@@ -363,6 +362,7 @@ class Bucket(S3Bucket):
                                 ' is required to access ACLs.</Details>') %
                                details)
                     if six.PY3:
+                        # All args to re.sub() must be of same type
                         details = details.encode('utf-8')
                     body = re.sub(ERROR_DETAILS_REGEX, details, body)
             raise self.connection.provider.storage_response_error(
--- a/boto/gs/bucket.py
+++ b/boto/gs/bucket.py
@@ -26,9 +26,6 @@ from __future__ import print_function
 import re
 import xml.sax
 
-from six import PY3
-from six.moves.urllib.parse import quote
-
 import boto
 from boto import handler
 from boto.resultset import ResultSet
@@ -46,6 +43,8 @@ from boto.s3.bucket import Bucket as S3Bucket
 from boto.utils import get_utf8_value
 from boto.compat import six
 
+quote = six.moves.urllib.parse.quote
+
 # constants for http query args
 DEF_OBJ_ACL = 'defaultObjectAcl'
 STANDARD_ACL = 'acl'
@@ -54,7 +53,7 @@ ENCRYPTION_CONFIG_ARG = 'encryptionConfig'
 LIFECYCLE_ARG = 'lifecycle'
 STORAGE_CLASS_ARG='storageClass'
 ERROR_DETAILS_REGEX_STR = r'<Details>(?P<details>.*)</Details>'
-if PY3:
+if six.PY3:
     ERROR_DETAILS_REGEX_STR = ERROR_DETAILS_REGEX_STR.encode('ascii')
 ERROR_DETAILS_REGEX = re.compile(ERROR_DETAILS_REGEX_STR)
 
@@ -363,7 +362,7 @@ class Bucket(S3Bucket):
                     details = (('<Details>%s. Note that Full Control access'
                                 ' is required to access ACLs.</Details>') %
                                details)
-                    if PY3:
+                    if six.PY3:
                         details = details.encode('utf-8')
                     body = re.sub(ERROR_DETAILS_REGEX, details, body)
             raise self.connection.provider.storage_response_error(
--- a/boto/gs/bucket.py
+++ b/boto/gs/bucket.py
@@ -19,6 +19,10 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
 # IN THE SOFTWARE.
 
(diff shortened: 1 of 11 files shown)


boto/gs/key.py
```diff
@@ -707,7 +713,12 @@ def set_contents_from_string(self, s, headers=None, replace=True,
         self.md5 = None
         self.base64md5 = None
 
-        fp = StringIO(get_utf8_value(s))
+        if six.PY3:
+            if isinstance(s, str):
+                s = s.encode('utf-8')
+            fp = io.BytesIO(s)
+        else:
+            fp = io.BytesIO(get_utf8_value(s))
```
walkerjoe: `io.BytesIO()` does not exist in Py2. We should change this to use `six.BytesIO`
Info: https://pythonhosted.org/six/#six.BytesIO

boto/utils.py
```diff
@@ -860,11 +864,15 @@ def notify(subject, body=None, html_body=None, to_string=None,
 
 
 def get_utf8_value(value):
-    if not six.PY2 and isinstance(value, bytes):
-        return value
+    # if not six.PY2 and isinstance(value, bytes):
+    #     return value
```
walkerjoe: Commented code should probably be removed.

walkerjoe (changes requested):

boto/gs/key.py
catleeball: Nice catch! Updated to six.BytesIO

boto/utils.py
catleeball: Got it, removed

catleeball:
Joe and I are currently investigating and fixing errors that Travis found. :construction_worker_woman:

boto/gs/bucket.py
```diff
@@ -40,14 +43,20 @@
...
+
 # constants for http query args
 DEF_OBJ_ACL = 'defaultObjectAcl'
 STANDARD_ACL = 'acl'
 CORS_ARG = 'cors'
 ENCRYPTION_CONFIG_ARG = 'encryptionConfig'
 LIFECYCLE_ARG = 'lifecycle'
 STORAGE_CLASS_ARG='storageClass'
-ERROR_DETAILS_REGEX = re.compile(r'<Details>(?P<details>.*)</Details>')
+ERROR_DETAILS_REGEX_STR = r'<Details>(?P<details>.*)</Details>'
```
houglum: Nit: If this is only used to create `ERROR_DETAILS_REGEX` below (i.e. it shouldn't be "publicly" discoverable by other modules), consider naming it with a leading underscore, `_ERROR_DETAILS_REGEX_STR`

boto/gs/bucket.py
```diff
@@ -353,6 +362,8 @@ def _get_xml_acl_helper(self, key_name, headers, query_args):
                     details = (('<Details>%s. Note that Full Control access'
                                 ' is required to access ACLs.</Details>') %
                                details)
+                    if six.PY3:
```
houglum, line 364: Probably worth adding a comment here mentioning *why* we do this -- because all arguments to `re.sub()` must be the same type.

catleeball:
Regarding the python 2.6 failure, it looks like one of boto's dependencies, [idna](https://github.com/kjd/idna) no longer supports Python 2.6 as of [this PR](https://github.com/kjd/idna/commit/2618ad17c86498d627ebb136143c2799ad48cbe3#diff-2eeaed663bd0d25b7e608891384b7298) due to 2.6's deprecation.

This error seems to occur in 2.6 for other recent PRs to Boto as well.

Boto team: just to check, does Boto still officially support deprecated Python versions including Python 2.6, 3.3, and 3.4?

Will continue investigating the other Travis errors 🔍

boto/gs/bucket.py
```diff
@@ -40,14 +43,20 @@
 from boto.utils import get_utf8_value
 from boto.compat import six
 
+quote = six.moves.urllib.parse.quote
```
houglum: IIUC, most of the compatibility stuff should be taken from `boto.compat`, rather than directly from the source `boto.compat` takes its aliases from (e.g. its vendored version of `six`).  Can we change this line to:

```
from boto.compat import quote
```

boto/gs/key.py
```diff
@@ -707,7 +713,12 @@ def set_contents_from_string(self, s, headers=None, replace=True,
         self.md5 = None
         self.base64md5 = None
 
-        fp = StringIO(get_utf8_value(s))
+        if six.PY3:
```
houglum, line 716: Looking at `get_utf8_value`, I don't think this does anything different in Python 3 than in 2.  I think we can revert this block of changes.

boto/utils.py
```diff
@@ -860,11 +864,12 @@ def notify(subject, body=None, html_body=None, to_string=None,
 
 
 def get_utf8_value(value):
-    if not six.PY2 and isinstance(value, bytes):
+    if six.PY3:
+        if isinstance(value, bytes):
+            return value.decode('utf-8')
```
houglum: Why is this being decoded into unicode? That's not what this method's name indicates that it does.

Did you mean to just write `value.decode('utf-8')`, without the return statement, so that this fails if the bytes can't be encoded using the utf-8 codec?

boto/utils.py
```diff
@@ -860,11 +864,12 @@ def notify(subject, body=None, html_body=None, to_string=None,
...
         return value
-
     if not isinstance(value, six.string_types):
-        value = six.text_type(value)
+        value = six.text_type(value).encode('utf-8')
```
houglum: This isn't necessary.  The next lines will detect that this is an instance of `text_type` and do the same encoding that's being done here.  This line can be reverted.

boto/utils.py
```diff
@@ -1096,3 +1101,81 @@ def parse_host(hostname):
         return hostname.split(']:', 1)[0].strip('[]')
     else:
         return hostname.split(':', 1)[0]
+
+
+def ttyprint(*objects, **kwargs):
+  """ A Python 2&3 compatible version of the print function.
```
houglum, line 1115: The indentation in this function is 2 spaces; it should be 4.

boto/s3/key.py
```diff
@@ -1549,7 +1553,7 @@ def _get_file_internal(self, fp, headers=None, cb=None, num_cb=10,
             cb(data_len, cb_size)
         try:
             for bytes in self:
-                fp.write(bytes)
+                boto.utils.ttyprint(bytes, file=fp, end='')
```
houglum, line 1556: The name on this method seems to be misleading -- we're not necessarily writing to a tty, correct? `fp` could refer to a file descriptor that isn't a terminal.  Maybe the method should be named `write_to_file_or_tty`?

boto/gs/bucket.py
catleeball: SGTM, done 👍

boto/s3/key.py
catleeball: Sounds good to me. What about `write_to_fd` or `write_to_file_descriptor`? A smidge shorter for the former, and file descriptor covers both, especially since `file=` should be able to take any kind of fd.

boto/gs/bucket.py
catleeball: Agreed, this is weird syntax, updated

boto/gs/bucket.py
catleeball: SGTM, done.

boto/utils.py
catleeball: I honestly don't know the rationale, this was pulled in from Ed's changes. Does changing line 869 to `value = value.decode('utf-8')` sound reasonable? I think you're right, the return statement looks like it was probably a typo or something.

boto/utils.py
catleeball: Sounds right to me, reverted.

boto/utils.py
catleeball: Updated to `value = value.decode('utf-8')` per above in most recent commit.

houglum: My issue with this change is that this method is called "get_utf8_value", meaning that the thing being returned should be a utf-8-encoded `bytes` object, not a unicode object.  The only reason that a `value.decode('utf-8')` line would be present would be to make sure that the bytes passed in could be decoded using UTF-8.

Given this confusion, I think it's worth also adding a comment to the docstring saying that the return type of this method is a `bytes` value.

boto/s3/key.py
houglum: Sounds good. Might even be worth revisiting the logic to see if we can break it up into something like:

```
if fd_is_tty():
  write_to_tty(data)
else:  # fd is a file
  write_to_file(data)
```

... if that's possible.  It's a bit tough to follow as-is.

catleeball: Yeah, it's due for a refactor, I'll do break it up a bit :+1:

catleeball:
Moved the commit for upgrading the six library to https://github.com/boto/boto/pull/3852

catleeball:
Moved travis update to https://github.com/boto/boto/pull/3853

boto/s3/key.py
catleeball: Since this is the single and only call in boto to ttyprint / ttywrite, I'm going to axe those functions from boto entirely and just do six.ensure_binary() here. The refactor to those functions I'll include in gsutil where those functions are used more widely

catleeball:
Split utils.py changes into https://github.com/boto/boto/pull/3854

catleeball:
Split mock storage integ test fix into https://github.com/boto/boto/pull/3855

catleeball:
Migrated remaining changes to https://github.com/boto/boto/pull/3856 closing this PR

PR #3855, opened 2019-03-19
Make mock storage integ test Python 2/3 compatible

catleeball:
This change was done by Ed Hodapp. It includes checking the type
of file descriptor and data and transforming data as appropriate
so that it can be written regardless of encoding.


houglum (approved):

houglum:
LGTM. @mfschwartz, could you merge this when you get a chance?

As noted in the previous PR, the Py2.6 TravisCI environment was already failing.

PR #3857, opened 2019-03-21
[Moved: see PR #3858] Revert integ test changes

catleeball:
This reverts commit 6521d009b740536b1ccba3c29c187f42c09e8a61.

This was one of the changes we recovered from Ed's VM. Although boto's tests pass with this change, gsutil's tests don't. Apologies on the testing gap.

PR #3858, opened 2019-03-21
Fix test failures introduced in #3855 and #3866.

houglum:
This is a superset of the proposed changes in https://github.com/boto/boto/pull/3857.

Also changes variable name so that it's not using the name of a built-in
type, i.e. `for key_bytes in self` instead of `for bytes in self`.

While the rollback here might affect Python 3 compatibility in gsutil, we can address
that later - our main focus is not breaking anything that was already working. These
fixes restore the previous functionality; I verified that all boto integration tests and
gsutil integration tests (using a local copy of boto that includes these changes) pass.

+ @catleeball for FYI

### DIFF
--- a/boto/s3/key.py
+++ b/boto/s3/key.py
@@ -392,7 +392,7 @@ class Key(object):
         By providing a next method, the key object supports use as an iterator.
         For example, you can now say:
 
-        for bytes in key:
+        for key_bytes in key:
             write bytes to a file or whatever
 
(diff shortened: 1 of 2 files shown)


houglum:
@mfschwartz Please merge these whenever you get the chance. In addition to the TravisCI tests passing (aside from Py 2.6), I also double-checked and ran all the integration tests (for both Boto and gsutil) to make sure everything is passing again.

PR #3861, opened 2019-03-27
Ensure strings can be utf-8 encoded

catleeball:
## Summary
We replaced the old method get_utf8_value with get_utf8able_str.
It attempts to make objects a unicode string that can by utf8 encoded.

## Testing
`https://github.com/boto/boto/pull/3861.diff` was applied to a local
copy of `gsutil/gslib/vendored/boto`. Integration tests ran successfully
on Linux using Python 2.7.15 with JSON and XML APIs.

### DIFF
--- a/boto/connection.py
+++ b/boto/connection.py
@@ -1105,8 +1105,10 @@ class AWSQueryConnection(AWSAuthConnection):
     def _required_auth_capability(self):
         return []
 
+
     def get_utf8_value(self, value):
         return boto.utils.get_utf8_value(value)
+
 
     def make_request(self, action, params=None, path='/', verb='GET'):
         http_request = self.build_base_http_request(verb, path, None,
--- a/boto/connection.py
+++ b/boto/connection.py
@@ -1106,6 +1106,15 @@ class AWSQueryConnection(AWSAuthConnection):
         return []
 
 
+    def get_utf8_value(self, value):
+        """This replaces public interface get_utf8_value with new implementation.
+        
+        The old implementation of get_utf8_value has been deprecated and replaced
+        with get_utf8able_str.
+        """
+        return boto.utils.get_utf8able_str(value)
+
+
     def make_request(self, action, params=None, path='/', verb='GET'):
         http_request = self.build_base_http_request(verb, path, None,
                                                     params, {}, '',
--- a/boto/connection.py
+++ b/boto/connection.py
@@ -1107,12 +1107,7 @@ class AWSQueryConnection(AWSAuthConnection):
 
 
     def get_utf8_value(self, value):
-        """This replaces public interface get_utf8_value with new implementation.
-        
-        The old implementation of get_utf8_value has been deprecated and replaced
-        with get_utf8able_str.
-        """
-        return boto.utils.get_utf8able_str(value)
+        return boto.utils.get_utf8_value(value)
 
 
     def make_request(self, action, params=None, path='/', verb='GET'):
--- a/boto/connection.py
+++ b/boto/connection.py
@@ -1105,8 +1105,6 @@ class AWSQueryConnection(AWSAuthConnection):
     def _required_auth_capability(self):
         return []
 
-    def get_utf8_value(self, value):
-        return boto.utils.get_utf8_value(value)
 
     def make_request(self, action, params=None, path='/', verb='GET'):
         http_request = self.build_base_http_request(verb, path, None,
--- a/boto/s3/bucket.py
+++ b/boto/s3/bucket.py
@@ -51,6 +51,7 @@ import re
 import base64
 from collections import defaultdict
 from boto.compat import BytesIO, six, StringIO, urllib
+from boto.utils import get_utf8able_str
 
 # as per http://goo.gl/BDuud (02/19/2011)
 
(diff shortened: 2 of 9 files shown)


catleeball:
@houglum Added `get_utf8able_str` function per your feedback in chat, thank you for helping me understand the complexities of encoding. :bowing_woman:

boto/s3/bucket.py
```diff
@@ -852,7 +853,7 @@ def copy_key(self, new_key_name, src_bucket_name,
             if isinstance(src_key_name, bytes):
                 src_key_name = src_key_name.decode('utf-8')
         else:
-            src_key_name = boto.utils.get_utf8_value(src_key_name)
+            src_key_name = get_utf8able_str(src_key_name)
```
houglum: I think this whole if/else statement can be replaced with `src_key_name = get_utf8able_str(src_key_name)`, right? The if block above is just doing what get_utf8able_str would do for the PY3 case anyway.

boto/utils.py
```diff
@@ -862,20 +862,6 @@ def notify(subject, body=None, html_body=None, to_string=None,
             boto.log.exception('notify failed')
 
 
-def get_utf8_value(value):
```
houglum: This was a public visible method (i.e. not named with a leading underscore) in the API for the `utils` module, meaning users of the Boto library might have been depending on it.

Given this, I'd say we should keep this method here, but add a docstring saying that it's deprecated and has been replaced with `get_utf8able_str` within the Boto library.

houglum (changes requested):

boto/connection.py
```diff
@@ -1105,8 +1105,6 @@ def __init__(self, aws_access_key_id=None, aws_secret_access_key=None,
     def _required_auth_capability(self):
         return []
 
-    def get_utf8_value(self, value):
```
houglum: Also add this back in.  Other parts of the codebase relied on this method:

```
<...>/boto/boto $ grep "get_utf8_value(" . -r
./mturk/connection.py:            value = connection.get_utf8_value(value)
./mturk/connection.py:            value = connection.get_utf8_value(value)
[...]
```

catleeball: Ah, good catch! Will re-add.

boto/utils.py
catleeball: Makes sense, will do!

boto/s3/bucket.py
catleeball: Looks like it! Cut out the if else block.

boto/connection.py
```diff
@@ -1106,6 +1106,15 @@ def _required_auth_capability(self):
         return []
 
 
+    def get_utf8_value(self, value):
+        """This replaces public interface get_utf8_value with new implementation.
+        
+        The old implementation of get_utf8_value has been deprecated and replaced
+        with get_utf8able_str.
```
houglum: Let's not change the implementation out from underneath people (meaning that it would still exist but behave differently all of sudden -- uh oh).  Can we just make this call `boto.utils.get_utf8_value` again?

catleeball:
Fixed all the comments you left @houglum , about to re-run tests with the new PR diff in my gsutil.

boto/connection.py
catleeball: Ah! Sure thing, I think I misread the prior comment. Will revert to the old implementation.

boto/utils.py
```diff
@@ -1101,6 +1101,41 @@ def parse_host(hostname):
...
+        # We want to return a unicode/str object.
+        if isinstance(s, six.text_type):
+            return s
+        if isinstance(s, six.binary_type):
+            s = s.decode('utf-8')
+            return s
+    raise TypeError('not expecting type "%s"' % type(s))
+
+
+def get_utf8_value(self, value):
```
catleeball: Reverting this to the old implementation

houglum: Same comment on this one -- internally, we can switch all the stuff in Boto to call `get_utf8able_str` instead.  But we don't want to change behavior for or break external users who were relying on this method and its existing behavior.  Thus, this should still call `boto.utils.get_utf8_value()`, rather than `get_utf8able_str`.

catleeball: Got that fixed! Github was giving me guff, I had to make another PR and merge it into this one's branch, but the diff should have all the feedback resolved now. :)

catleeball:
Re-ran tests from gsutil with the most recent set of changes:
✅ Linux Python 2.7.15 JSON
✅ Linux Python 2.7.15 XML

catleeball:
While gsutil is happy with get_utf8able method, boto unit tests aren't, particularly in the case where it's something that isn't a string but has the `__str__` property, like ints. I'll look more into that.

houglum (approved):
LGTM

catleeball:
In boto integ tests, mturk is upset, I must have mixed up one of the public interfaces, or messed up reverting get_utf8_value. I'll investigate that now.

catleeball:
Okay, now with mturk updated and values encoded before being hashed, boto tests appear happy. :)

houglum:
@mfschwartz Can you merge this when you get the chance? We've tested this PR against boto and gsutil integration tests, and all TravisCI tests are passing except for Py2.6 (which has been broken for a while and is unrelated to this PR)

PR #3880, opened 2019-09-04
Fix for Python 4: replace unsafe six.PY3 with PY2

hugovk:
We don't yet know if 3.10 or 4.0 will follow Python 3.9, but whichever it is, it will probably happen in [2020 when Python 3.9 reaches beta](https://www.python.org/dev/peps/pep-0596/#schedule) and work begins on Python 3.9+1.

There's some code which uses `six.PY3`:

```python
if six.PY3:
    print("Python 3+ code")
else:
    print "Python 2 code"
```

Where:

```python
PY3 = sys.version_info[0] == 3
```

When run on Python 4, this will run the Python 2 code! Instead, use `six.PY2`.

Found using https://github.com/asottile/flake8-2020:
```console
$ pip install -U flake8-2020
...
$ flake8 --select YTT
./boto/compat.py:60:4: YTT202 `six.PY3` referenced (python4), use `not six.PY2`
./boto/connection.py:657:36: YTT101 `sys.version[:3]` referenced (python3.10), use `sys.version_info`
./boto/connection.py:658:21: YTT101 `sys.version[:3]` referenced (python3.10), use `sys.version_info`
./boto/glacier/utils.py:128:8: YTT202 `six.PY3` referenced (python4), use `not six.PY2`
./boto/vendored/six.py:37:7: YTT201 `sys.version_info[0] == 3` referenced (python4), use `>=`
./boto/vendored/six.py:635:8: YTT203 `sys.version_info[1]` compared to integer (python4), compare `sys.version_info` to tuple
./tests/unit/glacier/test_utils.py:143:26: YTT202 `six.PY3` referenced (python4), use `not six.PY2`
./tests/unit/dynamodb/test_types.py:123:26: YTT202 `six.PY3` referenced (python4), use `not six.PY2`
./tests/unit/dynamodb/test_types.py:143:26: YTT202 `six.PY3` referenced (python4), use `not six.PY2`
./tests/integration/dynamodb/test_layer2.py:463:22: YTT202 `six.PY3` referenced (python4), use `not six.PY2`
```

---

See also:

* https://github.com/boto/boto3/pull/2256
* https://github.com/boto/botocore/pull/1933
* https://github.com/boto/s3transfer/pull/152

PR #3885, opened 2019-09-26
Issue #3544

cvesay:
* `boto\mws\connection.py`
    * Corrected String != Bytes assertion bug in `get_report`.  Based on suggested fix by @listingmirror.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,9 +1,9 @@
 ####
 boto
 ####
-boto 2.49.0
+boto 2.49.1
 
-Released: 11-July-2018
+Released: TBD
 
 .. image:: https://travis-ci.org/boto/boto.svg?branch=develop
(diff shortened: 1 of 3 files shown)


cvesay:
Updated the patch to include all additional updates made up through v2.49.0

cvesay:
https://github.com/boto/boto/pull/3880#issuecomment-527851751

cvesay:
Package can be installed as a dependency through setuptools.

```
install_requires=[
        'boto @ https://github.com/cvesay/boto/archive/develop.zip#egg=boto-2.49.1',
        .
        .
        .
]
```

cvesay:
Added some logic to return a XML body when requesting an XML report from MWS.  There was a bug with properly parsing an XML file, would only return the last element in the list

PR #3898, opened 2020-01-20
Fix collections.abc imports for Python 3.9

hugovk:
Since Python 3.3, `Mapping` etc. has been available in `collections.abc`.

Since Python 3.9, `Mapping` etc. is only available in `collections.abc`.

https://docs.python.org/3.9/whatsnew/3.9.html#removed

```pycon
Python 3.9.0a2+ (heads/master:ec007cb, Jan  5 2020, 12:34:22)
[Clang 10.0.1 (clang-1001.0.46.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import collections
>>> collections.Mapping
  File "<stdin>", line 1, in <module>
AttributeError: module 'collections' has no attribute 'Mapping'
>>> collections.abc.Mapping
  File "<stdin>", line 1, in <module>
AttributeError: module 'collections' has no attribute 'abc'
>>> import collections.abc
>>> collections.abc.Mapping
<class 'collections.abc.Mapping'>
>>>
```

Python 3.9 is already in alpha with full release due in October.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,9 @@ matrix:
       python: "3.7"
     - os: linux
       dist: xenial
+      python: "3.9-dev"
+    - os: linux
+      dist: xenial
       python: "pypy3.5"
   allow_failures:
     - python: "3.7"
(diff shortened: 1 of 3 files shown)


hugovk:
For a follow-up: nose also doesn't support Python 3.9 and is unmaintained, so `test/test.py` will need rewriting or redoing.

Issue #3906, opened 2020-06-25
Is boto supporting Signature Version 4 for SES?


PeterCui1991:
Recently got notification about SES signature version 3 will be deprecated on Oct 2020. But I didn't see any documentation about boto supporting Signature Version 4 for SES. Is there anyone who can help?

nolson-berry:
Researching, found this on SO: https://stackoverflow.com/questions/62757921/is-aws-boto-python-supporting-ses-signature-version-4

PeterCui1991:
@nolson-berry Thank you so much, yeah, that question was created by me on stackoverflow too.

shazi7804:
+1, Which version boto3 support Signature Version 4 for SES?

bwesen:
Hi, like the above posters, I have a legacy project on AWS and I keep getting spammed with these messages that my account uses Signature Version 3 on SES and we're using boto (not boto3). The project is legacy and I want to change as few things as possible to resolve this.

I didn't understand from the above - there are no versions of boto (not boto3) that supports Signature Version 4 on SES? SV4 was introduced before 2016.. 

Can you easily mix boto3 with boto in an old project (using Python 2.7) as a workaround, and only use boto3 for SES maybe?

kdaily:
Hi there,

This package is deprecated and no longer supported. Please review the deprecation notice for migration to `boto3`. Thanks!

https://github.com/boto/boto#deprecation-notice

PR #3916, opened 2021-01-21
Making SES use signature V4 for auth instead of signature V3

sleg-aws:
https://github.com/boto/boto/issues/3906


sleg-aws:
unfortunately this will never get merged, boto2 is deprecated and won't receive any updates.

PR #3946, opened 2022-10-26
Fixing SyntaxWarning: "is" with a literal in python3.8 and above

jiashuChen:
The python 3.8 compiler now produces a SyntaxWarning when identity checks (is and is not) are used with certain types of literals (e.g. strings, numbers).

These can often work by accident in CPython, but are not guaranteed by the language spec.

Changing from 'is' to '==' should be backward compatible as well


jiashuChen:
closing PR as I was informed that this package is deprecated, and we should always use boto3

